Issue #1, opened 2015-07-31
Is there a way to also support memrchr?


nicokoch:
I was wondering if it would be possible for this library to also support memrchr?
I can see this might be difficult because memrchr is not a posix-defined functionality. 

And another (kinda related) question: Does this crate work on all platforms (especially windows)?

BurntSushi:
> I was wondering if it would be possible for this library to also support memrchr?
I can see this might be difficult because memrchr is not a posix-defined functionality.

I would be OK adding it. IMO, we should provide an alternative when it's not available though.

> And another (kinda related) question: Does this crate work on all platforms (especially windows)?

Yes. This crate is used inside `regex` for some critical/common optimizations (which is tested on Windows).

nicokoch:
Thanks for the reply! I will create a PR in the near future. 
> we should provide an alternative when it's not available though.

Would something using `Iterator::rposition()` suffice?

BurntSushi:
> Would something using Iterator::rposition() suffice?

Absolutely. Either I or someone else can come along later and provide a more efficient implementation for systems that don't have `memrchr`. :-)

bluss:
This can be closed.

BurntSushi:
Thanks to @bluss! :)

PR #2, opened 2015-08-01
Implement safe interface to memrchr.

nicokoch:
This commit introduces the new function memrchr(), which is the reversed
counterpart to memchr(), and  it's corresponding testcases and benchmarks.

As discussed in #1. It looks live mingw [provides](https://github.com/dscho/msys/blob/master/msys/packages/coreutils/5.97/lib/memrchr.h) `memrchr()`, so a windows specific implementation is not necessary (not tested though, will check tomorrow). I am not really sure about other platforms though.

### DIFF
--- a/benches/bench.rs
+++ b/benches/bench.rs
@@ -24,3 +24,21 @@ fn libc_memchr(b: &mut test::Bencher) {
         assert!(memchr::memchr(needle, &haystack).is_none());
     });
 }
+
+#[bench]
+fn iterator_reversed(b: &mut test::Bencher) {
+    let haystack = bench_data();
+    let needle = b'a';
+    b.iter(|| {
+        assert!(haystack.iter().rposition(|&b| b == needle).is_none());
+    });
+}
+
+#[bench]
+fn libc_memrchr(b: &mut test::Bencher) {
+    let haystack = bench_data();
+    let needle = b'a';
+    b.iter(|| {
+        assert!(memchr::memchr(needle, &haystack).is_none());
+    });
+}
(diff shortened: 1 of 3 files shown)


BurntSushi:
cc @alexcrichton What are your thoughts on this? This commit adds `memrchr` which is not part of POSIX (while `memchr` is). Do we just have to go through each supported platform and check whether `memrchr` is provided?

Also, I think we need to make sure this is available with MSYS on Windows too I think? Not sure.

alexcrichton:
I'd recommend setting up AppVeyor to see if this works on Windows ([example config](https://github.com/alexcrichton/filetime/blob/master/appveyor.yml)), but a quick check of my local OSX system indicates that `memrchr` isn't available on OSX by default.

nicokoch:
So I just added #[cfg] gates and chose the defensive approach of only building the native version on linux.
We also have an appveyor.yml now (You just have to enable it for this project on the website).
You can see the builds for this commit [here](https://ci.appveyor.com/project/nicokoch/rust-memchr).

benches/bench.rs
```diff
@@ -24,3 +24,21 @@ fn libc_memchr(b: &mut test::Bencher) {
...
+        assert!(haystack.iter().rposition(|&b| b == needle).is_none());
+    });
+}
+
+#[bench]
+fn libc_memrchr(b: &mut test::Bencher) {
+    let haystack = bench_data();
+    let needle = b'a';
+    b.iter(|| {
+        assert!(memchr::memchr(needle, &haystack).is_none());
```
bluss: ah, typo -- should be mem**r**chr

BurntSushi:
Sorry for the delay in merging, but everything looks good now. It's up on crates.io in `0.1.4`. Thanks!

PR #3, opened 2015-08-19
Provide a faster fallback for memrchr

bluss:
There are patches in the pipeline for 16-bit pointer sizes and so on in the main rust tree, so that got me thinking about making sure it compiles no matter what the target pointer size is. Specialization is only for 32-bit and 64-bit platforms.

benchmark looks like this with the fallback used for memrchr, tested on 64-bit linux (fallback is normally not used in linux, but here it is).

```
test iterator          ... bench:      11,799 ns/iter (+/- 36) = 847 MB/s
test iterator_reversed ... bench:      11,781 ns/iter (+/- 29) = 848 MB/s
test libc_memchr       ... bench:         455 ns/iter (+/- 3) = 21978 MB/s
test libc_memrchr      ... bench:       2,236 ns/iter (+/- 12) = 4472 MB/s
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -172,13 +172,13 @@ mod fallback {
 
         while offset >= 2 * USIZE_BYTES {
             unsafe {
-                let mut u = *(ptr.offset(offset as isize - 2 * USIZE_BYTES as isize) as *const usize);
-                let mut v = *(ptr.offset(offset as isize - USIZE_BYTES as isize) as *const usize);
+                let u = *(ptr.offset(offset as isize - 2 * USIZE_BYTES as isize) as *const usize);
+                let v = *(ptr.offset(offset as isize - USIZE_BYTES as isize) as *const usize);
 
                 // break if there is a matching byte
-                u ^= repeated_x;
-                v ^= repeated_x;
-                if contains_zero_byte(u) || contains_zero_byte(v) {
+                let zu = contains_zero_byte(u ^ repeated_x);
+                let zv = contains_zero_byte(v ^ repeated_x);
+                if zu || zv {
                     break;
                 }
             }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,8 +8,8 @@ to the corresponding functions in `libc`.
 extern crate libc;
 
 use libc::funcs::c95::string;
-use libc::types::common::c95::c_void;
-use libc::types::os::arch::c95::{c_int, size_t};
+use libc::c_void;
+use libc::{c_int, size_t};
 
 /// A safe interface to `memchr`.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -8,8 +8,8 @@ to the corresponding functions in `libc`.
 extern crate libc;
 
 use libc::funcs::c95::string;
-use libc::types::common::c95::c_void;
-use libc::types::os::arch::c95::{c_int, size_t};
+use libc::c_void;
+use libc::{c_int, size_t};
```
bluss: I couldn't resist changing this. The long paths bug me.

BurntSushi: Hah, it wasn't intentional. I always get lost in the module hierarchy of `liblibc`. :-(

bluss:
Oh :laughing:  memrchr fallback is [faster](https://ci.appveyor.com/project/BurntSushi/rust-memchr/build/1.0.8/job/bddb7c3amyy7cdp3) than memchr itself on windows according to the appveyor tests..

I can of course provide the same algorithm in its forward direction, if wanted.

BurntSushi:
@bluss What the..? Really? If you want to, sure, that benchmark seems pretty convincing...

bluss:
Found an improvement.. new bench is

```
test libc_memrchr      ... bench:       1,974 ns/iter (+/- 26) = 5065 MB/s
```

just slightly better, but this one actually vectorized on my tested platform.

bluss:
I'll come back to memchr another day. This is fine to merge if you want.

For memchr, I wonder what the OS X situation looks like?

BurntSushi:
@bluss Thanks!

No clue about OS X. I have a Mac mini next to me that has been sitting in its box since December. Been meaning to get it set up to build/dist Rust programs...

nicokoch:
Just a heads up: I tested this implementation on my own windows box and @bluss's implementation is clearly faster than windows' memchr:
```
running 4 tests
test iterator          ... bench:       4,767 ns/iter (+/- 3,443) = 2097 MB/s
test iterator_reversed ... bench:       4,630 ns/iter (+/- 1,386) = 2159 MB/s
test libc_memchr       ... bench:       4,062 ns/iter (+/- 1,345) = 2461 MB/s
test libc_memrchr      ... bench:         870 ns/iter (+/- 230) = 11494 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured
```

BurntSushi:
@bluss I knew you were good... but I didn't know you were *that* good!

bluss:
wow! I want to thank llvm for the codegen :wink:

PR #4, opened 2015-08-22
Provide faster memchr for windows

bluss:
Follow up PR #3 with the same algorithm in forward search, manually implemented
memchr searching two usize at a time. This should be better than some libc implementations
on windows, so we enable this by default there.

I'm not very happy with the tangle of configuration conditions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "memchr"
 version = "0.1.5"  #:version
-authors = ["Andrew Gallant <jamslam@gmail.com>"]
+authors = ["Andrew Gallant <jamslam@gmail.com>", "bluss"]
 description = "Safe interface to memchr."
 documentation = "http://burntsushi.net/rustdoc/memchr/"
 homepage = "https://github.com/BurntSushi/rust-memchr"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -297,16 +297,44 @@ mod tests {
     }
 
     #[test]
-    fn qc_correct_reversed() {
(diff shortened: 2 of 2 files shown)


bluss:
I suspect based on googling that OS X's impl is similarly unimpressive. We just need to find somone to test it.

programble:
Benchmark results on Mac OS X 10.11 Beta (15A262e), Rust 1.4.0-nightly (8f1b0aa32 2015-08-21):

```
test iterator          ... bench:       4,123 ns/iter (+/- 1,108) = 2425 MB/s
test iterator_reversed ... bench:       3,772 ns/iter (+/- 1,058) = 2651 MB/s
test libc_memchr       ... bench:         154 ns/iter (+/- 30) = 64935 MB/s
test libc_memrchr      ... bench:         754 ns/iter (+/- 224) = 13262 MB/s
```

bluss:
Cool. Looks like OS X's got a solid vectorized memchr implementation already.

bluss:
Thanks for running this!

bluss:
@programble To explain, the libc_memchr scans the data so fast it is probably using something like SSE4.2 extensions. The memrchr result is for an implementation in pure rust, which is still a pretty good improvement compared to a naive loop at least. With this result we'll continue to use libc memchr for OS X.

bluss:
Ok inspired by something I saw from @kimundi, I guess a build script is the only way to set our own config flags. It would move the configuration choices out and keep it a bit cleaner.

BurntSushi:
Awesome! Thanks! :-)

PR #5, opened 2015-12-19
Use memrchr definition from libc

bluss:
libc 0.2.4 includes memrchr (for linux, netbsd, openbsd). Reportedly the fallback is still better for the bsds.


BurntSushi:
Sweet! Thanks.

PR #6, opened 2016-06-19
dependencies: require libc 0.2.4 as minimum

posborne:
Many crates depend on libc.  Having the dependency here require
exactly 0.2.4 causes problems when some other requires a version
newer than 0.2.4.  By sepcifying '^0.2.4' we allow any 0.2.x
version where `x > 4`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ name = "memchr"
 bench = false
 
 [dependencies]
-libc = "0.2.4"
+libc = "^0.2.4"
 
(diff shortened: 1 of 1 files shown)


bluss:
Those two requirements should be equivalent (0.2.4 is not exact, ^ is the default).

posborne:
Hmmm, yeah, you're right.  It turns out I am an idiot and the problem I was running into was just related to not running `cargo update` when moving to a new version of nix on one of my projects.

PR #10, opened 2016-10-08
Implement Iterator for memchr

cholcombe973:
So I partially implemented Iterator to see what you think.  I can add the trait for the other memchr2,3 and memrchr if you like this.  Some of the formatting of the file changed due to me having rustfmt that automatically runs on every save.  If it bothers you I can turn it off and fix the PR.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,8 +98,13 @@ impl<'a> DoubleEndedIterator for Memchr<'a> {
             Some(index) => {
                 // Move our internal position
                 self.haystack = self.haystack.split_at(index).0;
-                self.rposition = self.position + index + 1;
-                Some(self.rposition)
+                if self.rposition > 0{
+                    //memrchr has been used once already.
+                    self.rposition = self.rposition - index;
+                    return Some(self.rposition);
+                }else{
+                    return Some(self.position + index+1);
+                }
             }
             None => None,
         }
@@ -763,6 +768,22 @@ mod tests {
         assert_eq!(Some(15), second);
         assert_eq!(Some(10), third);
         assert_eq!(None, fourth);
+    }
+
+    #[test]
+    fn memrchr_iter(){
+        let haystack = b"aaaabaaaabaaaab";
+        let mut memchr_iter = Memchr::new(b'b', haystack);
+        let first = memchr_iter.next_back();
+        let second = memchr_iter.next_back();
+        let third = memchr_iter.next_back();
+        let fourth = memchr_iter.next_back();
+
+        assert_eq!(Some(15), first);
+        assert_eq!(Some(10), second);
+        assert_eq!(Some(5), third);
+        assert_eq!(None, fourth);
+
     }
 
     #[test]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,6 @@
-/*!
-This crate defines two functions, `memchr` and `memrchr`, which expose a safe interface
-to the corresponding functions in `libc`.
-*/
+//! This crate defines two functions, `memchr` and `memrchr`, which expose a safe interface
+//! to the corresponding functions in `libc`.
+//!
 
 #![deny(missing_docs)]
 #![allow(unused_imports)]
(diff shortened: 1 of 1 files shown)


bluss:
You can use one double ended iterator to cover both memchr and memrchr. Don't need to keep `position` (or rposition), just split & update the slice as you go.

cholcombe973:
@bluss thanks! Yeah I like that idea. I'll update the PR to implement double ended iterator.  I have an idea of how to update the slice as I go for iterator but I'm not sure how to make that work on next_back(). I'll give it some thought

bluss:
Remember a double ended iterator uses a range where you can pick from each end, I think it should work well.

bluss:
I mean, if you take a few elements off the back with `.next_back()`, those will not show up in `.next()`.

cholcombe973:
@bluss and @BurntSushi so I think this correct but it could use a double check.  Especially on the tests.  I'm not 100% certain the next_back() test is correct.

src/lib.rs
```diff
@@ -1,7 +1,6 @@
-/*!
-This crate defines two functions, `memchr` and `memrchr`, which expose a safe interface
-to the corresponding functions in `libc`.
-*/
+//! This crate defines two functions, `memchr` and `memrchr`, which expose a safe interface
+//! to the corresponding functions in `libc`.
+//!
```
BurntSushi: Could you not change the comment style please? :-)

src/lib.rs
```diff
@@ -52,6 +51,58 @@ fn repeat_byte(b: u8) -> usize {
...
+pub struct Memchr<'a> {
+    needle: u8,
+    // The haystack to iterate over
+    haystack: &'a [u8],
+    // The index
+    position: usize,
+}
+
+impl<'a> Memchr<'a> {
+    /// Create a new Memchr that's initalized to zero with a haystack
```
BurntSushi: I think the docs could be a little better. In particular, "initialized to zero" seems like an implementation detail that doesn't belong in the public API. What about this?

> Creates a new iterator that yields all positions of `needle` in `haystack`.

Similarly for the other iterators.

BurntSushi:
@cholcombe973 This largely looks like great work, thank you. :-) I have a couple of nits. Aside from that, do we want the iterator names to be `Memchr`, `Memchr2` and `Memchr3`? I think our naming conventions generally seem to prefer `Iter` if there's one primary iterator in a module. But there are three. Are `Iter`, `Iter2` and `Iter3` better?

BurntSushi:
@cholcombe973 Also, sorry, but since I merged @bluss's PR a few moments ago, you'll need to rebase before I can merge.

src/lib.rs
```diff
@@ -531,14 +658,61 @@ mod tests {
...
+    fn memchr_reverse_iter() {
+        let haystack = b"aaaabaaaab";
+        let mut memchr_iter = Memchr::new(b'b', haystack);
+        let first = memchr_iter.next();
+        let second = memchr_iter.next_back();
+        let third = memchr_iter.next();
+
+        assert_eq!(Some(5), first);
+        assert_eq!(Some(10), second);
+        assert_eq!(None, third);
```
BurntSushi: I believe this looks right. Could you add a third `b` to `haystack` and run another `next_back` call?

src/lib.rs
cholcombe973: Yeah that sounds great

src/lib.rs
cholcombe973: Yeah I think that 3rd `b` did the function in.  I don't think the output is correct.
```
let haystack = b"aaaabaaaabaaaab";
let mut memchr_iter = Iter::new(b'b', haystack);
let first = memchr_iter.next();
println!("first: {:?}", first);
let second = memchr_iter.next_back();
println!("second: {:?}", second);
let third = memchr_iter.next();
println!("third: {:?}", third);
let fourth = memchr_iter.next_back();
println!("fourth: {:?}", fourth);
//output
first: Some(5)
second: Some(15)
third: Some(20)
fourth: None
```

bluss:
@BurntSushi Isn't the convention that the iterator creation method is named identically with the iterator itself? And vice versa. I think the memchr based names seem better than Iter and so on.

src/lib.rs
```diff
@@ -52,6 +51,58 @@ fn repeat_byte(b: u8) -> usize {
...
+
+impl<'a> DoubleEndedIterator for Memchr<'a> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        let search_result = memrchr(self.needle, &self.haystack);
+        match search_result {
+            Some(index) => {
+                // Move our internal position
+                self.haystack = self.haystack.split_at(index).0;
+                self.position = self.position + index + 1;
+                Some(self.position)
```
bluss: "position" is the front position, and it shouldn't move in next_back. The iterator element should be `position + index` here, shouldn't it? (A.k.a postion + haystack.len() after haystack is updated).

If that doesn't seem right, add more quickchecks. :D Exhaustive testing is better than single tests. (So yay quickcheck).

BurntSushi:
@bluss I guess I'm just thinking about it as `memchr::Iter`, kind of like how we have `slice::Iter`. But... I guess that is produced by `slice.iter()`, and in this case, we have `Memchr::new` producing the iterator `Memchr`. So I guess I agree with you. Let's keep the names as they in the PR.

cholcombe973:
I think the splitting is what is making memrchr confused.  I'll see what I can do about it

cholcombe973:
Alright so I got it working.  I couldn't figure out how to do this without adding an rposition field :-/

src/lib.rs
```diff
@@ -52,6 +52,60 @@ fn repeat_byte(b: u8) -> usize {
...
+}
+
+impl<'a> DoubleEndedIterator for Memchr<'a> {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        let search_result = memrchr(self.needle, &self.haystack);
+        match search_result {
+            Some(index) => {
+                // Move our internal position
+                self.haystack = self.haystack.split_at(index).0;
+                self.rposition = self.position + index + 1;
```
BurntSushi: Should this be `self.rposition = self.rposition + index + 1`?

BurntSushi:
@cholcombe973 Great work! Looks good to me. :-) Thanks so much!!

If there's a way to avoid the `rposition` field then we can improve on it later.

cholcombe973:
Awesome!  Thanks

PR #12, opened 2017-01-03
Add optional no_std support

mbrubeck:
This still uses std by default, for compatibility with versions of Rust older than 1.6.0.  If compatibility with old versions is not a concern, this could be simplified to remove the optional feature and instead set `no_std` unconditionally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ rust:
   - beta
   - nightly
 script:
+  - cargo build --verbose --no-default-features
   - cargo build --verbose
   - cargo test --verbose
   - cargo doc
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,12 @@ license = "Unlicense/MIT"
 name = "memchr"
 bench = false
 
+[features]
+default = ["use_std"]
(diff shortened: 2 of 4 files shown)


BurntSushi:
@mbrubeck Thanks so much for this! Good news too: I just recently released `1.0.0` of this crate, and its minimum supported version is Rust 1.12, so we can use `no_std` unconditionally.

... however, what happens if we do want to use `std` some day? I hope to goodness we don't, but if we did, that would be a breaking change, right?

mbrubeck:
Note: Pushed a minor change to fix the build on Rust 1.12.0 and 1.13.0.

> ... however, what happens if we do want to use `std` some day? I hope to goodness we don't, but if we did, that would be a breaking change, right?

Yes, that would be another possible reason to keep `use_std` as a default feature.

BurntSushi:
@mbrubeck Yes, for that reason, I would like to keep the feature present. More concretely, once we get SIMD, I expect this crate to get some more attention. Ideally, that itself will only require `core` as well, but I can't quite see that far into the future.

BurntSushi:
`memchr 1.0.1` is on crates.io. Thanks again!

PR #14, opened 2017-02-11
Add Memchr*::haystack, to get the current haystack being iterated over.

burtonageo:
This can be useful in situations where you need to check the bytes relative to the position of the haystack.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,6 +83,30 @@ impl<'a> Memchr<'a> {
             rposition: 0,
         }
     }
+
+    /// Get a reference to the haystack currently being iterated over.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # fn main() {
+    /// # use memchr::Memchr;
+    ///
+    /// let haystack = "Hello, world!\n, How are you\r\n";
+    /// let mut memchr = Memchr::new(b'\n', &haystack.as_bytes());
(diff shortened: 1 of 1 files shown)


burtonageo:
@BurntSushi Is this commit something you want for rust-memchr? It would be very useful for the `Lines` iterator in the ascii crate

BurntSushi:
I guess this should probably be added for all of the iterators, right? Including `memchr3`?

I think some of the iterators in `std` have similar methods. What names do they use?

In principle, I'm fine with a change like this.

bluss:
I think the doc is a bit ambiguous — like, is it clear that it's not the original haystack?

src/lib.rs
```diff
@@ -83,6 +83,30 @@ impl<'a> Memchr<'a> {
...
+    /// # Example
+    ///
+    /// ```rust
+    /// # fn main() {
+    /// # use memchr::Memchr;
+    ///
+    /// let haystack = "Hello, world!\n, How are you\r\n";
+    /// let mut memchr = Memchr::new(b'\n', &haystack.as_bytes());
+    /// while let Some(pos) = memchr.next() {
+    ///     if let Some(&b'\r') = memchr.haystack().get(pos - 1) {
```
bluss: This example doesn't do what it appears to do at first glance -- because `.haystack()` is changing during iteration.

Issue #13, opened 2017-01-16
Off by one index in Memchr iterator


matt2xu:
It seems surprising to me that when using the Memchr iterator, the position that is returned when calling .next() is the position of the needle + 1. Looking at the tests of the crate, this is expected behavior. Could you please explain why it is this way? I could then make a PR with additional documentation and an example to show usage of the Memchr (and Memchr2/3) iterator.

BurntSushi:
Uh... That seems surprising to me and sounds like a bug. cc @cholcombe973

cholcombe973:
Yeah I seem to recall the problem was if I didn't return the needle position plus one it would get stuck in a infinite loop.

BurntSushi:
*sigh* OK. Returning the index + 1 doesn't seem right to me at all. I really wish I caught that.

This has to be fixed. The question now is whether this requires a `2.0.0` release. It probably does.

cholcombe973:
It seemed right at the time :-/. I think I was using it to split strings.

matt2xu:
Any update on this? It would be nice to be zero-based IMO.

BurntSushi:
Fixed in #18. Note that we decided it would be best to do a memchr 2 release for this change.

PR #16, opened 2017-06-16
Remove dead code

cbreeden:
It's not possible for `Memchar.rposition > 0` so this branch is dead.

All tests pass.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -70,7 +70,6 @@ pub struct Memchr<'a> {
     haystack: &'a [u8],
     // The index
     position: usize,
-    rposition: usize,
 }
 
 impl<'a> Memchr<'a> {
@@ -80,7 +79,6 @@ impl<'a> Memchr<'a> {
             needle: needle,
             haystack: haystack,
             position: 0,
-            rposition: 0,
         }
     }
 }
@@ -109,13 +107,7 @@ impl<'a> DoubleEndedIterator for Memchr<'a> {
             Some(index) => {
                 // Move our internal position
                 self.haystack = self.haystack.split_at(index).0;
-                if self.rposition > 0 {
-                    //memrchr has been used once already.
-                    self.rposition = self.rposition - index;
-                    Some(self.rposition)
-                } else {
-                    Some(self.position + index+1)
-                }
+                Some(self.position + index + 1)
             }
             None => None,
         }


bluss:
If all the tests still pass, then the iterator or tests were already broken. Removing rposition doesn't make any sense; the iterator is not double ended (DEI) this way. A double ended iterator has two ends, and you can alternate taking elements from either side. Taking from one side does not alter the other side, unless you reach the point where the ends meet.

Sketch about the concept of a DEI.

```rust
[a b c d] start
[. b c d] .next() -> a
[. b c .] .next_back() -> d
[. b . . ] .next_back() -> c
[. . . . ] .next() -> b
[. . . . ] .next() or .next_back() -> None
```

bluss:
I may be kneejerking a bit too hard there, since we have the slice to keep the progress. I'll read this more carefully now. At least we have a nice diagram for DEI...

BurntSushi:
Is #13 related?

bluss:
The code in this PR looks correct to me and it passes the extra quickcheck test I tested it with.

The off by one error looks simple to fix, code-wise, but it's orthogonal.

bluss:
The extra qc test for Memchr's DEIness is not especially elegant, but I did write it so why not share it.

```rust
    // take items from a DEI, taking from the front for each true and back for each
    // false. Return a vector with the concatenation of the fronts and the
    // reverse of the backs.
    fn double_ended_take<I, J>(mut iter: I, take_side: J) -> Vec<I::Item>
        where I: DoubleEndedIterator,
              J: Iterator<Item=bool>,
    {
        let mut found_front = Vec::new();
        let mut found_back = Vec::new();

        for take_front in take_side {
            if take_front {
                if let Some(pos) = iter.next() {
                    found_front.push(pos);
                } else {
                    break;
                }
            } else {
                if let Some(pos) = iter.next_back() {
                    found_back.push(pos);
                } else {
                    break;
                }
            };
        }

        let mut all_found = found_front;
        all_found.extend(found_back.into_iter().rev());
        all_found
    }

    #[test]
    fn qc_correct_memchr_iter() {
        fn prop(find: u8, data: Vec<u8>, mut take_side: Vec<bool>, take_last: bool) -> bool {
            // nonempty sequence of booleans
            take_side.push(take_last);

            let iter = Memchr::new(find, &data);
            let all_found = double_ended_take(iter, take_side.iter().cycle().cloned());

            // Note: Indices off by one
            all_found.iter().cloned().eq(
                data.iter().enumerate().filter_map(|(i, elt)| if *elt == find { Some(i + 1) } else { None }))
        }
        quickcheck::quickcheck(prop as fn(_, _, _, _) -> _);
    }
```

cbreeden:
@bluss thanks for looking into this.  I understand the initial reaction of "this just seems wrong".  I had exactly the same reaction, and that's what eventually led me to this PR.  Also reading `memrchr` as `memchr` led me astray for a good 20 minutes as well :)

cbreeden:
Would you like me to add some more tests?  Perhaps the quickcheck test that @bluss wrote above.

bluss:
I was inspired to make the other PR, it has more tests. It's the part with breaking changes though (fixing the indexing paradigm)

bluss:
We can of course steal tests from that PR if we want to do that and wait with the indexing change.

PR #18, opened 2017-08-05
Fix iterators to use 0-based indexing

### LINKED ISSUE #13: Off by one index in Memchr iterator

bluss:
This fixes Memchr, Memchr2, Memchr3 to use 0-based indexing.

- PR #16 needs to be merged first.
- The iterators are tested by using `.iter().enumerate()` as reference. This should ensure the iterators follow the conventional 0-based indexing correctly.
- Includes misc change that cuts down the number of crates compiled to run tests.
- Includes size hint for the iterators

**Note:** This is a breaking change.

Fixes #13

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ use_std = ["libc/use_std"]
 libc = { version = "0.2.18", default-features = false }
 
 [dev-dependencies]
-quickcheck = "0.4.1"
+quickcheck = { version = "0.4.1", default-features = false }
 
 [profile.test]
 opt-level = 3
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -991,13 +991,13 @@ mod tests {
             // test that the size hint is within reasonable bounds
             let needle = 0;
             let mut iter = Memchr::new(needle, &data);
-            let real_total = data.iter().filter(|&&elt| elt == needle).count();
-            let mut found = 0;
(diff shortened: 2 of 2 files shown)


bluss:
@BurntSushi hi! Alternatively, we can deprecate these iterators, not fix them, and redirect people to another crate.. and we put them there.

Main problem with that idea is that I don't see myself using the iterators, so I'd not be the ideal maintainer.

BurntSushi:
@bluss Thanks for this. Sorry it's taken me so long to look at it. What do you think of just releasing a memchr 2?

bluss:
Memchr 2 seems unproblematic since this crate is not exactly used as public dependency often.

bluss:
It almost looks like I got lazy in the PR, the tests are not exactly consistent and exhaustive, even if they are pretty good. Just let me know what you are missing.

bluss:
Don't squash merge :smile:

BurntSushi:
I rebased and fixed conflicts locally, and merged this to master. Also, `memchr 2` is now on crates.io!

BurntSushi:
Thanks @bluss so much for fixing this. :-) <3

PR #20, opened 2017-10-11
Make linking to libc optional

Amanieu:
When compiling with `no_std` for systems with no C library we want to avoid linking to the `libc` crate. This PR makes `libc` an optional feature which is enabled by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,11 +19,11 @@ name = "memchr"
 bench = false
 
 [features]
-default = ["use_std"]
-use_std = ["libc/use_std"]
+default = ["use_std", "libc"]
+use_std = ["libc", "libc/use_std"]
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


Amanieu:
Any update on this? We need this change for the [cstr_core](https://crates.io/crates/cstr_core) crate, which allows C strings to be used in bare-metal environments.

bluss:
Is there any risk of build break due to this? It looks like previous versions were not really no std at all, since libc pulls in libstd.

Was it ever right to call libc memchr in this crate when using the "no-std" configuration?

Amanieu:
In most no_std use cases you don't want to link to libc at all, usually because your target system doesn't have a libc to link to (kernel or bare-metal).

BurntSushi:
All right, this is on crates.io in `memchr 1.0.2`. Thanks!

PR #24, opened 2018-01-01
fallback: fix variable name in docstrings

hdhoang:
I intend to make the same PR for `src/libcore/slice/memchr.rs` in rust-lang/rust. If you are OK with this, I will r=you there.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -424,7 +424,7 @@ mod fallback {
         contains_zero_byte, repeat_byte,
     };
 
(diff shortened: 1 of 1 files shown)


hdhoang:
All of travis' jobs failed with

```
oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"
```

That's troubling.

BurntSushi:
Thanks! And yeah, Travis has been on the fritz the past couple days.

PR #30, opened 2018-03-03
clippy: Fix needless borrow warnings.

waywardmonkeys:
`self.haystack` here is already a reference, so no need to take
a reference to it.


BurntSushi:
@waywardmonkeys Thanks for these fixes! If you're going to submit a lot of them, could you please queue them up into one PR with each in its own commit?

waywardmonkeys:
These are the last ones for now for `memchr`!

waywardmonkeys:
I have 2 open against `tabwriter` though!

PR #39, opened 2018-10-27
exclude CI files

ignatenkobrain:


BurntSushi:
Thanks! This change is on crates.io in `memchr 2.1.1`.

PR #42, opened 2018-11-24
Use static with AtomicPtr instead of static mut for ifunc

bluss:
There's a faint rumbling in the distance, that static mut is to be
avoided. It seems to be a problem when we take *mutable* references to
the static, and this code didn't.

Still, port the code to using a regular static with an atomic pointer.

I have been admiring this code :heart_eyes:, first to understand it, then to remember
to copy it when I need it. Looking forward to the discussion about
soundness.

### DIFF
--- a/src/x86/mod.rs
+++ b/src/x86/mod.rs
@@ -1,5 +1,5 @@
 use std::mem;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicPtr, Ordering};
 
 use fallback;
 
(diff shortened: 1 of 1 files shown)


bluss:
Oh I ran the memchr1/rust/tiny benchmarks and did not find a difference. I ran the benchmark first *with the PR* then *without*.

The second *without* output says there was no difference outside the noise. But I'm unsure which bench you've been using for the ifunc code — I'll run the ones you think are best for that.

```
memchr1/rust/tiny/never time:   [27.711 ns 27.820 ns 27.959 ns]                                     
                        thrpt:  [2.2984 GiB/s 2.3099 GiB/s 2.3190 GiB/s]
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe

memchr1/rust/tiny/rare  time:   [37.767 ns 37.880 ns 38.049 ns]                                    
                        thrpt:  [1.6889 GiB/s 1.6964 GiB/s 1.7015 GiB/s]
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe

memchr1/rust/tiny/uncommon                                                                            
                        time:   [81.522 ns 81.905 ns 82.435 ns]
                        thrpt:  [798.25 MiB/s 803.41 MiB/s 807.18 MiB/s]
Found 19 outliers among 100 measurements (19.00%)
  8 (8.00%) high mild
  11 (11.00%) high severe

memchr1/rust/tiny/common                                                                            
                        time:   [167.95 ns 168.34 ns 168.86 ns]
                        thrpt:  [389.69 MiB/s 390.90 MiB/s 391.80 MiB/s]
Found 12 outliers among 100 measurements (12.00%)
  6 (6.00%) high mild
  6 (6.00%) high severe
```

src/x86/mod.rs
```diff
@@ -40,15 +42,15 @@ macro_rules! ifunc {
                 } else {
                     fallback::$name as usize
                 };
-            let slot = unsafe { &*(&FN as *const _ as *const AtomicUsize) };
+            let slot = unsafe { &*(FN.get() as *const AtomicUsize) };
```
eddyb: I think the solution is to use `AtomicPtr<()>` (or something else instead of `()`, but it doesn't really matter) instead of `AtomicUsize`.
Then, you might be able to put `AtomicPtr<()>` in the `static` itself, which *is the correct thing to do*.

`static mut` and `static`+`SyncUnsafeCell` are *approximations* of the real `static`+synchronized abstraction.

bluss: Ah that's perfect. Fn to usize cast was not allowed in statics, but fn to raw pointer is. Thanks!

bluss:
Updated to use `AtomicPtr<()>` like @eddyb suggested.

I/you can squash the PR when you've seen it.

eddyb:
@bluss Can you also change the PR title?

BurntSushi:
Thanks @bluss and @eddyb! Sorry for taking so long to get around to this, but I wanted to scrutinize this carefully before merging.

The benchmark that's probably relevant for this PR is `memchr1/rust/tiny/never` (or any of the `rust/tiny/never` benchmarks), particularly in comparison with `memchr1/libc/tiny/never`. I tried them as well and I couldn't see any difference from before/after this PR.

PR #45, opened 2019-01-04
ci: test on big endian

BurntSushi:
There is a hypothesis that the memchr rewrite from a while back is
getting things wrong on big endian, as shown here:
https://github.com/BurntSushi/ripgrep/issues/1144

As a result, we start testing in CI on a big endian architecture.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,11 @@ matrix:
     - os: linux
       rust: 1.13.0
       env: TARGET=x86_64-unknown-linux-gnu
+    # For testing big-endian.
+    - env: TARGET=mips64-unknown-linux-gnuabi64 CROSS=1
+      rust: stable
+      services: docker
+      sudo: required
 script: ci/script.sh
 branches:
   only:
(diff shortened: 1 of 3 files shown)


BurntSushi:
This doesn't seem to have uncovered any bugs. Upon looking more at the code, I can't actually spot anywhere that has endian specific code outside of the SIMD implementations, which can assume little endian since they only work on x86 CPUs. Still, it seems good to continue testing on a big endian platform now that the work is done.

.travis.yml
```diff
@@ -28,6 +28,11 @@ matrix:
     - os: linux
       rust: 1.13.0
       env: TARGET=x86_64-unknown-linux-gnu
+    # For testing big-endian.
+    - env: TARGET=mips64-unknown-linux-gnuabi64 CROSS=1
+      rust: stable
+      services: docker
+      sudo: required
 script: ci/script.sh
```
lzutao: `sudo` keyword is deprecated.

PR #49, opened 2019-02-12
trim dependences, make libc optional, fix no_std bug

### LINKED ISSUE #48: Clarity with regard to `--no-default-features` leading to `fallback::memchr`
First and foremost, thank you for the crate Andrew.

I’ve been employing it for a while, and I’ve recently made one of the libraries I’m working on `no_std`. Coincidentally while doing so, I altered my benchmarking habits, and never got hold of the fact `--no-default-features` leads to `fallback::memchr` due to `std::is_x86_feature_detected`, and only discovered the change via `cargo asm`.

What do you think of extending the README‘s `no_std` section to mention the otherwise “silent” requirement? 

(I am now using `std_detect` with an extern `no_std` memchr. Considering you work on that too, I assume you’ll want it to be stable before adopting it.)

BurntSushi:
This PR consists of several small tweaks and fixes:

* Use SSE2 routines even when `use_std` is disabled. Fixes #48.
* Simplify conditional compilation and remove `cfg-if` dependency.
* Don't use libc by default. Enabling the libc feature re-enables it. Since this should only be a matter of performance, we don't treat it as a breaking change.
* Rollup #42 into this PR, which uses AtomicPtr for the ifunc trick. cc @bluss


BurntSushi:
This PR is on crates.io in `memchr 2.2.0`.

PR #55, opened 2019-08-14
Support target_pointer_width less than 32

dflemstr:
My use-case for this is using the library for an avr target, and it seemed quite easy to get the fallback working for that use-case.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -25,11 +25,6 @@ instead of one. Similarly for `memchr3`.
 #![deny(missing_docs)]
 #![doc(html_root_url = "https://docs.rs/memchr/2.0.0")]
 
-// Supporting 16-bit would be fine. If you need it, please submit a bug report
-// at https://github.com/BurntSushi/rust-memchr
-#[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
-compile_error!("memchr currently not supported on non-32 or non-64 bit");
-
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -25,11 +25,6 @@ instead of one. Similarly for `memchr3`.
 #![deny(missing_docs)]
 #![doc(html_root_url = "https://docs.rs/memchr/2.0.0")]
 
-// Supporting 16-bit would be fine. If you need it, please submit a bug report
-// at https://github.com/BurntSushi/rust-memchr
-#[cfg(not(any(target_pointer_width = "32", target_pointer_width = "64")))]
-compile_error!("memchr currently not supported on non-32 or non-64 bit");
-
 #[cfg(feature = "use_std")]
```
timotree3: It seems more apt to keep this check and instead have it include the two new widths. That way if 128 bit targets or something like that become available, we give a better error message.

```suggestion
#[cfg(not(any(target_pointer_width = "8", target_pointer_width = "16", target_pointer_width = "32", target_pointer_width = "64")))]
compile_error!("memchr currently only supports 8, 16, 32, and 64 bit targets");
#[cfg(feature = "use_std")]
```

BurntSushi:
Thanks! I merged this in #65, but removed support for 8 bit pointers. I'm not sure Rust even supports that, and I'd want to give a careful audit to make sure it was kosher.

PR #60, opened 2019-09-25
Clarify bitshift amounts in sse2 pseudocode

timotree3:
I was very confused when I read this code because the bitshift amounts overlapped
so heavily and then are or-ed together. I believe the intention was to mean a number in bytes,
this changes it to a number in bits as bitshifts usually are.

### DIFF
--- a/src/x86/sse2.rs
+++ b/src/x86/sse2.rs
@@ -25,7 +25,7 @@ pub unsafe fn memchr(n1: u8, haystack: &[u8]) -> Option<usize> {
     // While this routine is very long, the basic idea is actually very simple
     // and can be expressed straight-forwardly in pseudo code:
     //
-    //     needle = (n1 << 15) | (n1 << 14) | ... | (n1 << 1) | n1
+    //     needle = (n1 << 120) | (n1 << 112) | ... | (n1 << 8) | n1
     //     while i <= haystack.len() - 16:
     //       // A 16 byte vector. Each byte in chunk corresponds to a byte in
     //       // the haystack.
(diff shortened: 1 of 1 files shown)


BurntSushi:
I'm not so sure about that. Shifts in SIMD code are not always expressed in bit amounts, but are sometimes expressed in terms of lanes: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_srli_epi16&expand=5506

I'd almost prefer leaving it as it is, and _maybe_ add a comment noting that they are byte shifts.

timotree3:
Fair enough, it's your choice. I'm very surprised that that's true and I would imagine most other programmers would be too.

Is there a better example of the lane amounts? In the link you posted it looks like the shift amount argument to the function is in bits to shift within each u16 and the number next to`>>` operator is a shift amount on a u16 that gets verified as less than 15. Seems like they're all in bits to me.

BurntSushi:
Oh, sorry, I linked to the wrong intrinsic. You're right. PALIGNR is what I had in mind: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_alignr_epi8&expand=257

I'll noodle on this.

timotree3:
I found a good example: https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm_srli_si128&expand=5533

When shifting an entire u128, the shift amount is expressed in bytes, which matches your point about lanes. Okay shall I change this PR to just add a comment?

BurntSushi:
Yeah I'd just add a comment noting it for now.

timotree3:
That CI failure was caused by https://github.com/rust-lang/cargo/issues/7427 
Want to re run it somehow? Seems rather unnecessary given that the changes total to a single comment.

PR #62, opened 2019-10-21
Expand documentation for `memr?chr[23]`

### LINKED ISSUE #61: memchr2_iter and memchr3_iter do not properly advance slice position
Hi @BurntSushi I'm the author of Artichoke Ruby. We met on Twitter.

Playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7f18fe26ae4413e95f1e350d77f86b28

The `iter_next!` macro hard codes how much to advance the haystack position by.

https://github.com/BurntSushi/rust-memchr/blob/1ec5ecce03c220c762dd9a8b08f7a3d95522b765/src/iter.rs#L17

This means that `memchr*_iter` functions on more than one byte incorrectly scan. For example, this code outputs 2 when it should output 1:

```rust
extern crate memchr; // 2.2.1

fn main() {
    let haystack = b"abcdefghijklmnopqrstuvwxyz";
    println!("{}", memchr::memchr2_iter(b'a', b'b', haystack.as_ref()).count());
}
```

timotree3:
Resolves #61 (in the sense that the misunderstanding leading to that issue will be prevented in the future)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -160,7 +160,28 @@ pub fn memchr(needle: u8, haystack: &[u8]) -> Option<usize> {
     }
 }
 
-/// Like `memchr`, but searches for two bytes instead of one.
+/// Like `memchr`, but searches for either of two bytes instead of just one.
+///
+/// This returns the index corresponding to the first occurrence of `needle1`
+/// or the first occurrence of `needle2` in `haystack` (whichever occurs
+/// earlier), or `None` if neither one is found.
+///
+/// While this is operationally the same as something like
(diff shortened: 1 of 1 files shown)


timotree3:
The CI failure seems spurious. It says that `archive.ubuntu.org` could not be reached. Seems like the Travis computer lost internet connection temporarily.

timotree3:
@BurntSushi have you gotten a chance to review this yet? It's not super urgent so I understand if you haven't. :)

PR #67, opened 2020-02-12
Support MIRI by using `naive` implementations

### LINKED ISSUE #51: Support miri
The crate does not seem to work well with miri at the moment for multiple reasons. One reason is that the crate uses the x86 / C implementations over the fallback implementation, which miri doesn't support. But even when cfg'ing out those implementations via cfg(miri), the fallback implementation uses a lot of bit math on pointers, which miri also doesn't like.

ldesgoui:
Closes #51
Some reordering was introduced in `cfg` attributes to make it consistent

```rust
if a {
    ...
} else if all(b, c) {
    ...
} else if d {
    ...
} else {
    ...
}
```
becomes
```rust
#[cfg(a)]
...
#[cfg(all(
    b,
    c,
    not(a),
))]
...
#[cfg(all(
   d,
   not(all(b, c, a)),
))]
...
#[cfg(all(
    not(d),
    not(all(b, c)),
    not(a),
))]
...
```
It highlights which component of the conditional is important/introduced by putting them first

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -53,7 +53,7 @@ mod iter;
 mod naive;
 #[cfg(test)]
 mod tests;
-#[cfg(all(target_arch = "x86_64", memchr_runtime_simd))]
+#[cfg(all(not(miri), target_arch = "x86_64", memchr_runtime_simd))]
 mod x86;
 
 /// An iterator over all occurrences of the needle in a haystack.
@@ -128,7 +128,13 @@ pub fn memrchr3_iter(
 /// ```
 #[inline]
 pub fn memchr(needle: u8, haystack: &[u8]) -> Option<usize> {
-    #[cfg(all(target_arch = "x86_64", memchr_runtime_simd))]
+    #[cfg(miri)]
+    #[inline(always)]
+    fn imp(n1: u8, haystack: &[u8]) -> Option<usize> {
+        naive::memchr(n1, haystack)
+    }
+
+    #[cfg(all(target_arch = "x86_64", memchr_runtime_simd, not(miri)))]
     #[inline(always)]
     fn imp(n1: u8, haystack: &[u8]) -> Option<usize> {
         x86::memchr(n1, haystack)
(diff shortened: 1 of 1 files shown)


ldesgoui:
Woops, looking into breakage

src/lib.rs
```diff
@@ -128,25 +128,33 @@ pub fn memrchr3_iter(
...
-    #[cfg(all(
-        not(memchr_libc),
-        not(all(target_arch = "x86_64", memchr_runtime_simd))
-    ))]
+    #[cfg(not(all(
+        miri,
+        target_arch = "x86_64",
+        memchr_runtime_simd,
+        memchr_libc
+    )))]
```
BurntSushi: I think this is your issue. I think you want:

```
#[cfg(all(
    not(memchr_libc),
    not(all(target_arch = "x86_64", memchr_runtime_simd)),
    not(miri),
))]
```

src/lib.rs
```diff
@@ -128,25 +128,33 @@ pub fn memrchr3_iter(
...
+    #[cfg(all(target_arch = "x86_64", memchr_runtime_simd, not(miri)))]
     #[inline(always)]
     fn imp(n1: u8, haystack: &[u8]) -> Option<usize> {
         x86::memchr(n1, haystack)
     }
 
     #[cfg(all(
         memchr_libc,
-        not(all(target_arch = "x86_64", memchr_runtime_simd))
+        not(all(miri, target_arch = "x86_64", memchr_runtime_simd))
```
BurntSushi: When using similar cfg knobs, could you make sure they are in the same order? e.g., I'd put `miri` last here, in order to match `all(target_arch = "x86_64", memchr_runtime_simd, not(miri))` above.

src/lib.rs
ldesgoui: Yep, `!A && !(B && C)` is `(!A && !B) || (!A && !C)` :man_facepalming:

BurntSushi:
This is on crates.io in `memchr 2.3.1`.

ldesgoui:
Thanks for the crate, the help and the responsiveness! Much appreciated

PR #70, opened 2020-02-12
Fix compilation error introduced in #67, increase test coverage against this issue

### LINKED ISSUE #68: Compilation Issue on rust 1.41.0
```rust
error[E0428]: the name `imp` is defined multiple times
   --> .cargo/registry/src/github.com-1ecc6299db9ec823/memchr-2.3.1/src/lib.rs:148:5
    |
139 |     fn imp(n1: u8, haystack: &[u8]) -> Option<usize> {
    |     ------------------------------------------------ previous definition of the value `imp` here
...
148 |     fn imp(n1: u8, haystack: &[u8]) -> Option<usize> {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `imp` redefined here
    |
    = note: `imp` must be defined only once in the value namespace of this block
```

ldesgoui:
Closes #68 
Apologies once again for the inconveniences

### DIFF
--- a/src/tests/miri.rs
+++ b/src/tests/miri.rs
@@ -1,0 +1,19 @@
+// Simple tests using MIRI
+
+use crate::{memchr, memchr2, memchr3, memrchr, memrchr2, memrchr3};
+
+#[test]
+fn test_with_miri() {
+    assert_eq!(memchr(b'a', b"abcda"), Some(0));
+    assert_eq!(memchr(b'z', b"abcda"), None);
+    assert_eq!(memchr2(b'a', b'z', b"abcda"), Some(0));
+    assert_eq!(memchr2(b'z', b'y', b"abcda"), None);
+    assert_eq!(memchr3(b'a', b'z', b'b', b"abcda"), Some(0));
+    assert_eq!(memchr3(b'z', b'y', b'x', b"abcda"), None);
+    assert_eq!(memrchr(b'a', b"abcda"), Some(4));
+    assert_eq!(memrchr(b'z', b"abcda"), None);
+    assert_eq!(memrchr2(b'a', b'z', b"abcda"), Some(4));
+    assert_eq!(memrchr2(b'z', b'y', b"abcda"), None);
+    assert_eq!(memrchr3(b'a', b'z', b'b', b"abcda"), Some(4));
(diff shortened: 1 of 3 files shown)


src/tests/miri.rs
```diff
@@ -0,0 +1,19 @@
...
+    assert_eq!(memchr2(b'z', b'y', b"abcda"), None);
+    assert_eq!(memchr3(b'a', b'z', b'b', b"abcda"), Some(0));
+    assert_eq!(memchr3(b'z', b'y', b'x', b"abcda"), None);
+    assert_eq!(memrchr(b'a', b"abcda"), Some(4));
+    assert_eq!(memrchr(b'z', b"abcda"), None);
+    assert_eq!(memrchr2(b'a', b'z', b"abcda"), Some(4));
+    assert_eq!(memrchr2(b'z', b'y', b"abcda"), None);
+    assert_eq!(memrchr3(b'a', b'z', b'b', b"abcda"), Some(4));
+    assert_eq!(memrchr3(b'z', b'y', b'x', b"abcda"), None);
+}
```
BurntSushi: Out of curiosity, why are these needed? Why can't miri just use the normal test suite?

(Sorry if this is a stupid question, I haven't actually used miri yet!)

BurntSushi:
All righty, let's try this again! :) This PR is on crates.io in `memchr 2.3.2`.

src/tests/miri.rs
ldesgoui: MIRI is magnitudes slower (100x, maybe even 1000x/10000x) than binaries built in debug mode
I tried running the test suite, the first test would show up as complete relatively quickly but no other test would complete for another few minutes
I assume there are timeout limits on GitHub actions, hence the short test, I suppose that the existing test suite could work just fine, I didn't really have the time to stick around :)

BurntSushi: Ah gotya. Yeah the tests check all the possible different offsets and alignments. This gets quite big because of the 256 bit vectors in the AVX version. It's probably possible to merge the test suites and just run fewer iterations when miri is enabled. I'll look into that when I get a chance!

PR #75, opened 2020-09-09
Run libc benchmarks on non-x86_64 archs as well

Count-Count:
I am trying to add aarch64 neon simd implementations (no promises though) and for that I need to compare it to the libc implementation which is already using neon instructions.

I removed the x86_64 guards in the benchmark code and it seems to be working fine.

### DIFF
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -19,9 +19,7 @@ path = "src/bench.rs"
 [dependencies]
 criterion = "0.2"
 memchr = { version = "*", path = ".." }
-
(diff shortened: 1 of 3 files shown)


BurntSushi (approved):
Hmmm yeah, I can't quite remember why I set it up this way.

> I am trying to add aarch64 neon simd implementations (no promises though)

If you do decide you want to do this, please open an issue and discuss the high level implementation path you're trying to take. The purpose of this is to ensure you aren't doing wasted work and that what you are doing will be mergeable. Thanks.

Issue #57, opened 2019-08-28
Compiler error when building on custom x86_64 no_std target


kevinaboos:
I ran into this error when trying to add the `cstr_core` crate to my no_std OS project, which depends on `memchr` and disables its default features.

Building in debug mode:
```
LLVM ERROR: Do not know how to split this operator's operand!
```
Building in release mode:
```
error: Could not compile `memchr`.

Caused by:
  process didn't exit successfully: `rustc --crate-name memchr /home/kevin/.cargo/registry/src/github.com-1ecc6299db9ec823/memchr-2.2.1/src/lib.rs --color always --crate-type lib --emit=dep-info,metadata,link -C opt-level=3 -C codegen-units=1 -C metadata=ebf07e953eacc227 -C extra-filename=-ebf07e953eacc227 --out-dir /home/kevin/my_os/target/x86_64-my_os/release/deps --target x86_64-my_os -L dependency=/home/kevin/my_os/target/x86_64-my_os/release/deps -L dependency=/home/kevin/my_os/target/release/deps --cap-lints allow --emit=obj -C debuginfo=2 -C code-model=large -C relocation-model=static -D unused-must-use -Z merge-functions=disabled -Z share-generics=no --sysroot /home/kevin/.xargo --cfg memchr_runtime_simd --cfg memchr_runtime_sse2 --cfg memchr_runtime_sse42 --cfg memchr_runtime_avx` (signal: 11, SIGSEGV: invalid memory reference)
```

For reference and in case it matters, my compiler target .json file is:
```
{
  "llvm-target": "x86_64-unknown-none-gnu",
  "data-layout": "e-m:e-i64:64-f80:128-n8:16:32:64-S128",
  "linker-flavor": "gcc",
  "target-endian": "little",
  "target-pointer-width": "64",
  "target-c-int-width": "32",
  "arch": "x86_64",
  "os": "none",
  "features": "-mmx,-sse,+soft-float",
  "disable-redzone": true,
  "panic": "abort"
}
```
My rustc version is `rustc 1.38.0-nightly (78ca1bda3 2019-07-08)`.

I've never encountered an error like this before, so I'm not sure what else to say. If more information is needed, I am happy to provide it.

PR #77, opened 2020-10-27
Only enable SIMD on targets which feature SSE

mwkroening:
I just ran into issue #57, which is caused by compiling this crate for targets that do not support SSE instructions. This is common for operating system kernels.

I don't consume this crate directly but through a dependency. It would be nice to have this crate compile without having to manually specify `memchr_disable_auto_simd`.

This PR makes the build script check if the target supports SSE and disables SIMD optimizations if not.

What do you think?

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -5,12 +5,19 @@ fn main() {
     enable_libc();
 }
 
-// This adds various simd cfgs if this compiler supports it.
+// This adds various simd cfgs if this compiler and target support it.
 //
 // This can be disabled with RUSTFLAGS="--cfg memchr_disable_auto_simd", but
 // this is generally only intended for testing.
+//
+// On targets which don't feature SSE2, this is disabled, as LLVM wouln't know
+// how to work with SSE2 operands. Enabling SSE4.2 and AVX on SSE2-only targets
+// is not a problem. In that case, the fastest option will be chosen at
+// runtime.
 fn enable_simd_optimizations() {
-    if is_env_set("CARGO_CFG_MEMCHR_DISABLE_AUTO_SIMD") {
+    if is_env_set("CARGO_CFG_MEMCHR_DISABLE_AUTO_SIMD")
+        || !target_has_feature("sse2")
+    {
         return;
     }
     println!("cargo:rustc-cfg=memchr_runtime_simd");
@@ -58,4 +65,10 @@ fn is_feature_set(name: &str) -> bool {
 
 fn is_env_set(name: &str) -> bool {
     env::var_os(name).is_some()
+}
+
+fn target_has_feature(feature: &str) -> bool {
+    env::var("CARGO_CFG_TARGET_FEATURE")
+        .map(|features| features.contains(feature))
+        .unwrap_or(false)
 }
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -10,7 +10,11 @@ fn main() {
 // This can be disabled with RUSTFLAGS="--cfg memchr_disable_auto_simd", but
 // this is generally only intended for testing.
 fn enable_simd_optimizations() {
-    if is_env_set("CARGO_CFG_MEMCHR_DISABLE_AUTO_SIMD") {
+    if is_env_set("CARGO_CFG_MEMCHR_DISABLE_AUTO_SIMD")
+        || !env::var("CARGO_CFG_TARGET_FEATURE")
+            .map(|features| features.contains("sse"))
+            .unwrap_or(false)
```
BurntSushi, line 16: Maybe move this conditional to a helper function with a sentence or two of docs explaining why we're gating SIMD on this?

BurntSushi (changes requested):
Hmmm... OK, so the high level problem here is that memchr is being compiled in an environment where SSE isn't supported, and that causes compilation to fail. I agree that that is a problem that we should fix, although I would _really_ prefer that we get a reproduction. #57 had issues with that too. Is there anyway I can actually test this fix and maybe add it to CI?

As for the solution in this PR, I have two concerns:

1. Is `CARGO_CFG_TARGET_FEATURE` actually set even when not explicitly set by the end user? If not, then this approach would disable SIMD optimizations in the vast majority of cases and thus would be a non-starter.
2. Do we need to check for SSE2, SSE42 and AVX too?

mwkroening:
Thanks for the quick reply!

I added a simple custom target without SSE and a CI job, building the project for this target (requires `nightly` and `rust-src`). It fails without and succeeds with the fix.

Regarding your concerns:

1. `CARGO_CFG_TARGET_FEATURE` is an [environment variable set by cargo](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts). For `x86_64-unknown-linux-gnu` it currently is `fxsr,sse,sse2`. For the custom target it is not set.

2. Testing different feature configurations I found that we only need SSE**2** support in order to compile the project. This allows for choosing SSE4.2 or AVX at runtime, if available.

build.rs
mwkroening: Done!

BurntSushi (approved):
All righty, LGTM! Thank you so much. I did check this out locally too, and tested it on Rust 1.28 (the MSRV) and all looks good. Thanks!

BurntSushi:
This PR is on crates.io in `memchr 2.3.4`.

mwkroening:
Thanks a lot!

BurntSushi:
@mkroening I ended up removing the CI test added in this PR. I did it because it was [failing in CI](https://github.com/BurntSushi/memchr/actions/runs/8030404672/job/21937751272). I did not understand the error message and I don't have the time to debug it.

I think going forward, one of the following three things should happen:

1. Support for testing this in CI needs to become more robust.
2. Its failure modes need to be better understandable by me.
3. We don't test it at all and instead rely on patches from stakeholders to fix things if they break. (This is the new status quo.)

PR #80, opened 2021-01-04
chore: Update libc and criterion crates to latest

stuartcarnie:
This PR updates the criterion crate to the latest, as the previous 0.2.x version no longer works with versions of Rust after 1.47. Specifically, running `cargo bench` would result in panics within Rayon.


BurntSushi (approved):
Thanks!

PR #84, opened 2021-06-13
Add wasm simd support

alexcrichton:
This commit adds simd acceleration support to the `memmem` module. This
is added with the freshly-stabilized support from rust-lang/rust#86204.
This mostly just cribs off the generic simd support for 128-bit types
built for sse, copying bits and pieces of code here and there. Some
refactoring happened internally to help reduce duplication where
possible.

I ran some initial benchmarks with the `memmem/krate/*` regex and a
hacked up single-threaded version of criterion. Some [initial
comparisons][compare] using Wasmtime as a runtime do indeed show a lot
of improvements, but there are indeed some slowdowns as well.

[compare]: https://gist.github.com/alexcrichton/6a72e682e7b6d505ade605359fbe3f2d

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
         - stable
         - stable-32
         - stable-mips
+        - wasm
         - beta
         - nightly
         - macos
@@ -62,6 +63,10 @@ jobs:
         - build: win-gnu
           os: windows-2019
           rust: stable-x86_64-gnu
+        - build: wasm
+          os: ubuntu-18.04
+          rust: stable-x86_64-gnu
+          wasm: true
     steps:
     - name: Checkout repository
       uses: actions/checkout@v1
@@ -81,6 +86,16 @@ jobs:
         cargo install --git https://github.com/rust-embedded/cross
         echo "CARGO=cross" >> $GITHUB_ENV
         echo "TARGET=--target ${{ matrix.target }}" >> $GITHUB_ENV
+    - name: Download Wasmtime
+      if: matrix.wasm
+      run: |
+        rustup target add wasm32-wasi
+        echo "CARGO_BUILD_TARGET=wasm32-wasi" >> $GITHUB_ENV
+        echo "RUSTFLAGS=-Ctarget-feature=+simd128" >> $GITHUB_ENV
+        curl -LO https://github.com/bytecodealliance/wasmtime/releases/download/v0.32.0/wasmtime-v0.32.0-x86_64-linux.tar.xz
+        tar xvf wasmtime-v0.32.0-x86_64-linux.tar.xz
+        echo `pwd`/wasmtime-v0.32.0-x86_64-linux >> $GITHUB_PATH
+        echo "CARGO_TARGET_WASM32_WASI_RUNNER=wasmtime run --wasm-features simd --" >> $GITHUB_ENV
     - name: Show command used for Cargo
       run: |
         echo "cargo command is: ${{ env.CARGO }}"
--- a/src/memmem/prefilter/mod.rs
+++ b/src/memmem/prefilter/mod.rs
@@ -1,8 +1,10 @@
 use crate::memmem::{rarebytes::RareNeedleBytes, NeedleInfo};
 
 mod fallback;
-#[cfg(all(target_arch = "x86_64", memchr_runtime_simd))]
+#[cfg(memchr_runtime_simd)]
 mod genericsimd;
+#[cfg(all(not(miri), target_arch = "wasm32", memchr_runtime_simd))]
+mod wasm;
 #[cfg(all(not(miri), target_arch = "x86_64", memchr_runtime_simd))]
 mod x86;
 
@@ -89,6 +91,21 @@ pub(crate) type PrefilterFnTy = unsafe fn(
     haystack: &[u8],
     needle: &[u8],
 ) -> Option<usize>;
+
+// If the haystack is too small for SSE2, then just run memchr on the
+// rarest byte and be done with it. (It is likely that this code path is
+// rarely exercised, since a higher level routine will probably dispatch to
+// Rabin-Karp for such a small haystack.)
+#[cfg(memchr_runtime_simd)]
+fn simple_memchr_fallback(
+    _prestate: &mut PrefilterState,
+    ninfo: &NeedleInfo,
+    haystack: &[u8],
+    needle: &[u8],
+) -> Option<usize> {
+    let (rare, _) = ninfo.rarebytes.as_rare_ordered_usize();
+    crate::memchr(needle[rare], haystack).map(|i| i.saturating_sub(rare))
+}
 
 impl PrefilterFn {
     /// Create a new prefilter function from the function pointer given.
@@ -269,7 +286,6 @@ impl PrefilterState {
 /// This only applies to x86_64 when runtime SIMD detection is enabled (which
 /// is the default). In general, we try to use an AVX prefilter, followed by
 /// SSE and then followed by a generic one based on memchr.
-#[cfg(all(not(miri), target_arch = "x86_64", memchr_runtime_simd))]
 #[inline(always)]
 pub(crate) fn forward(
     config: &Prefilter,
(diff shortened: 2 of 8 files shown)


alexcrichton:
I'll note that this is largely just a first stab, I think there's more simd acceleration in the `memchr` module but it looked much larger and less generic so it'll take some more time to do something like that. In any case I wanted to put this up for some initial feedback and whether this seems like the right direction to go in. Additionally if there's more benchmarking desired I'm happy to do so as well!

alexcrichton:
I hope it's ok to ping you again @BurntSushi, but if you're busy and/or would prefer to not land this, just let me know!

src/memmem/prefilter/mod.rs
```diff
@@ -280,20 +296,30 @@ pub(crate) fn forward(
...
+            return unsafe { Some(PrefilterFn::new(x86::sse::find)) };
+        }
     }
-    if cfg!(memchr_runtime_sse2) {
-        // SAFETY: x86::sse::find only requires the sse2 feature, which is
-        // guaranteed to be available on x86_64.
-        return unsafe { Some(PrefilterFn::new(x86::sse::find)) };
+    #[cfg(all(not(miri), target_arch = "wasm32", memchr_runtime_simd))]
+    {
+        if true {
```
BurntSushi, line 319: What's this `if true` for?

src/memmem/prefilter/mod.rs
```diff
@@ -306,28 +332,6 @@ pub(crate) fn forward(
...
-        return None;
-    }
-    let (rare1_rank, _) = rare.as_ranks(needle);
-    if rare1_rank <= MAX_FALLBACK_RANK {
-        // SAFETY: fallback::find is safe to call in all environments.
-        return unsafe { Some(PrefilterFn::new(fallback::find)) };
-    }
-    None
-}
-
```
BurntSushi, line 330: I assume this was unused?

BurntSushi (approved):
Nice! This overall looks great. And hopefully this makes it easier to add other arches too in the future.

Meta question: is it possible for `wasmtime` itself to just use this crate? (Or implement a better `memmem` itself?)

So I think my main concern here is unfortunately MSRV. I don't treat MSRV bumps as semver breaking changes, but I do generally like to stay conservative for widely used crates like this one. I imagine this bumps the MSRV, although, it looks like CI never ran...?

I think the MSRV thing either means waiting to land this (bummer) or adding version sniffing to `build.rs` that only enables this on a new enough version of Rust. I'd be willing to support that to unlock this change. (But no new deps please. Probably copy one of @dtolnay's build.rs files.)

And we should get CI running. Not sure why this PR didn't run.

BurntSushi:
Also, with respect to benchmarks, those look great. The slowdowns are generally what I'd expect I think. At least in theory, it would be quite surprising to see improvements across the board. The benchmarks are designed to put pressure on implementation choices that trade off perf in some cases for others.

src/memmem/prefilter/mod.rs
alexcrichton: Oh this was to prevent a lint from firing about dead code after this statement, I'll leave a comment

src/memmem/prefilter/mod.rs
alexcrichton: Oh I ended up folding this into the "generic" `forward` above and having internal `#[cfg]` about specialized implementations to avoid duplicating this definition in wasm (and also duplicating the fallback code)

alexcrichton:
Ok I think I fixed CI in https://github.com/BurntSushi/memchr/pull/84/commits/0d8dfc0433fe7dfb5da35d7292b81eb22b2146d0 (and I see now you've approved the CI runs as well, hurray me being a new contributor!)

> is it possible for wasmtime itself to just use this crate? 

I'm not quite sure what you mean by this, can you elaborate?

> So I think my main concern here is unfortunately MSRV

No worries! I'm happy to fixup anything that needs working. From the CI file it looks like it's 1.41.1, and to confirm is the goal that basically this should work on wasm by default with 1.41.1?

BurntSushi:
> I'm not quite sure what you mean by this, can you elaborate?

Yeah it might be an ill-formed question as a result of me knowing almost nothing about wasm haha. I probably have the abstraction layers totally confused to the point of it being a category error. But the question was provoked by your benchmarks. Presumably you're comparing the implementation in this crate with _something_. So I guess, what is that _something_? Where does it live? If it lives inside a runtime written in Rust (for example), then it seems plausible that it could be replaced with the code in this crate. Or maybe not?

> No worries! I'm happy to fixup anything that needs working. From the CI file it looks like it's 1.41.1, and to confirm is the goal that basically this should work on wasm by default with 1.41.1?

Yeah, because I think this _did_ work on Rust 1.41.1 with wasm? Although, maybe it's okay to have a more liberal MSRV policy there if the ecosystem is moving faster in wasm-land. Since that's the path of least resistance, let's just go that route until/if someone complains. But I think we should at least wait until it hits stable probably?

If MSRV on x86 remains unchanged (which I suspect is the case) then that's great.

alexcrichton:
Oh, heh, right! I should explain where the numbers came from... But nah nothing to do with Wasmtime in the sense of code in Wasmtime itself. The benchmark numbers before/after were this crate compiled with/without simd support (e.g. `-Ctarget-feature=+simd128` as a compilation flag or not). Wasmtime is interpreting the input as an opaque wasm blob and doesn't have a special `memmem` routine or anything like that. The result of this PR should be that if you compile a wasm binary with simd support (e.g. pass `-Ctarget-feature=+simd128`) then the binary will do simd things to get faster searching. Otherwise it'll continue to use the same fallback that exists today (and exhibit the "before" numbers in my gist above)

For MSRV, I'll leave that up to you. I don't think that this PR should affect any non-wasm targets (and if it does that's a bug I'm happy to fix). For wasm I don't think it actually affects MSRV unless you enable the simd feature, but the simd feature won't be stable until 1.54.0 (to be released tomorrow, what a coincidence I thought about this again today). If you'd like though I can implement support that forcibly disables simd support unless the rustc version is >= 1.54.0 on wasm.

I'm also find waiting until tomorrow to land if you'd prefer wasm simd support hit stable :)

BurntSushi:
Ooooooooo, gotya. That makes more sense.

OK, so my plan here is:

* Wait for 1.54 to come out (today).
* Try this out for myself on my local machine, and in particular, run the benchmarks to get experience with it.
* Run the normal x86 benchmarks just to check that there are no serious regressions on that side. (I don't expect there to be.)

And then I think it's good to go!

Also, I created https://github.com/BurntSushi/memchr/issues/87 for tracking the port of memchr to a more "generic" implementation. That would enable adding a memchr impl for WASM by "just" adding another shim that calls the generic version with a WASM vector type.

alexcrichton:
For running the benchmarks in wasm it's a bit tricky because criterion out-of-the-box doesn't work on wasm32-wasi (what I was using for running in `wasmtime`). It uses rayon in a few places and its dependency, `plotters`, equates wasm32 with "use wasm-bindgen for the web".

I had two local patches to get things working:

* [`plotters`](https://gist.github.com/alexcrichton/f23d91dd03ec313a127b7f5b70331cc1)
* [`criterion`](https://gist.github.com/alexcrichton/496cc24477587e8c627be588f7a7fbbd)

basically I made things assume single-threaded and not use web things to get things running in `wasmtime`. FWIW you can probably do web things today since I think at least Chrome & Firefox are shipping wasm simd support.

Otherwise though my command line to execute the wasm benchmarks was:

```
../wasmtime/target/release/wasmtime --mapdir target::./target -- no-simd.wasm --bench 'memmem/krate/*' --warm-up-time 1 --save-baseline no-simd
```

(searching in my bash history). I think I ran `cargo +nightly bench --no-run` and then copied the binary to `simd.wasm` and `no-simd.wasm`. The `--mapdir` business was so criterion had a place to write all of its results.

---

Also, if you're busy, I'm happy to look into the refactorings of #87 to get wasm supported with memchr as well as memmem. I skimmed the various modules and it looked like a simple enough port to the `Vector` trait with perhaps a method or two added. If you've got time though and feel inspired to take this on I'm happy to defer!

One question I had though reviewing the code. You've got an [`sse42` module](https://github.com/BurntSushi/memchr/blob/master/src/memchr/x86/sse42.rs) but I don't think it's listed in `x86/mod.rs`, so I don't think it's used? I'm not sure if that's intentional or whether it was in prototype phase and never got finished.

alexcrichton:
I was triggered into thinking about this again today, mind if I ping again @BurntSushi! I'm happy to help out with anything necessary to deal with this of course

.github/workflows/ci.yml
```diff
@@ -60,6 +63,10 @@ jobs:
         - build: win-gnu
           os: windows-2019
           rust: stable-x86_64-gnu
+        - build: wasm
+          os: ubuntu-18.04
+          rust: beta-x86_64-gnu # waiting for wasm simd to hit stable
```
BurntSushi, line 68: I think this can be switched to stable now?

BurntSushi (changes requested):
All righty! Thanks for the ping. The thing that was holding me up here was that I wanted to run the benchmark suite locally just to be sure there weren't any regressions. So far, nothing seems to stick out to me beyond some variance. (I'll publish this run eventually.) So I think this is good to go with one small change (switch CI to stable).

Also, could you squash this all down into one commit? (Or I can do it.)

Thanks!

alexcrichton:
Sure thing, done now! (squashed, switched to stable, and updated Wasmtime as well)

BurntSushi (approved):
w00t!

BurntSushi:
Urk. Sorry, I flubbed the benchmarks. Redoing them now. Once that's done, I'll put out a `memchr 2.5.0` release.

alexcrichton:
Oh no rush! I had a branch somewhere at some point that did most of the "other half" of this crate in porting it to the same style as what's here. I can try to clean that up to post it here so wasm gets a benefit from all the algorithms here instead of just some of them

BurntSushi:
Yeah that would be great! Thank you. :-)

BurntSushi:
@alexcrichton So sorry about the delay here, but I've _finally_ published this patch in `memchr 2.5.0`. I finally got around to running the benchmark suite both before and after this change (which takes 2.5 hours total! wow!) and it looks like there are no major differences.

alexcrichton:
No worries at all, and thanks again @BurntSushi!

PR #89, opened 2021-08-17
Add metadata to build as a dep of libstd

alexcrichton:
This commit adds the necessary `Cargo.toml` directives which enables
this crate to build with the standard library. This introduces some
optional off-by-deault dependencies which are only activated (and
overwritten) when built as part of the standard library.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,8 @@
 name: ci
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


BurntSushi (approved):
Thanks!

BurntSushi:
This PR is on crates.io in `memchr 2.4.1`.

PR #95, opened 2021-09-22
doc: find_iter, rfind_iter return non-overlapping matches

evanj:
The documentation for FindIter and FindRevIter already include this
detail. However, it was not immediately clear to me from the
documentation what find_iter and rfind_iter do about overlapping
matches. Add "non-overlapping" to the description.


evanj:
Thanks!

PR #118, opened 2023-02-09
Runtime customization of byte-frequency table

sentrip:


BurntSushi:
Yeah when you make changes, for a small PR like this, just keep everything in one commit. You can force push.

BurntSushi:
Hmmm can you submit another PR with the default type parameter approach? I think I'd still prefer that route if it can be made to work. Just wanted to look at it and play with it myself.

sentrip:
Yes I will upload it now, and thanks for the tip!

PR #120, opened 2023-02-13
Add compile time cpu feature detection when `no_std`

MarcusGrass:
Using `avx` instructions makes a big difference, being able to use them no-std would be cool. 
This PR attempts to add them by doing that feature detection at compile time if running `no_std`.  

The feature selection is a bit convoluted but it's supposed to be exclusive if I've got this right:
1. We only import `avx` on std, or if we've got `avx2` instructions.
2. We select `avx` on `no_std` if  we have `avx2` instructions.
3. We select `sse2` on `no_std` if we have `sse2` instructions but not `avx2` instructions.
4. We fallback on `no_std` if we don't have `avx2` or `sse2` instructions.

It's fairly difficult to tell if it's working, what I did was benchmark `std`, then I benchmarked with memchr `no-default-features`, then I benchmarked `RUSTFLAGS='-C target-cpu=native'` with `no-default-features` on a cpu that has `avx2`. 
The first and third bench was about on par, whereas the second had about half the throughput.

Edit: Woops, just saw https://github.com/BurntSushi/memchr/pull/106, but from reading the comments, this might be a good compromise.  
The use case is essentially micro optimization with dubious motivation.

### DIFF
--- a/src/memchr/x86/mod.rs
+++ b/src/memchr/x86/mod.rs
@@ -2,7 +2,7 @@ use super::fallback;
 
 // We only use AVX when we can detect at runtime whether it's available, which
 // requires std.
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", all(memchr_runtime_avx, target_feature = "avx2")))]
 mod avx;
 mod sse2;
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks! This PR is pretty simple so I think I'm likely to take it, but could you say what you're using this for? In particular, the use case for `no_std` on `x86_64`.

MarcusGrass:
Sure, I'm prototyping a [minimal x86/aarch64 std-library for linux](https://github.com/marcusGrass/tiny-std), it makes most simple CLI applications extremely small in comparison to using std, and in my still minimal testing, a bit faster. I'd like to use memchr since the Linux-APIs are c-centric so to make sure input is valid searching for null-terminators on user-provided input is often required.  

It's a circle where the reason for a tiny-std is to make tiny apps, and my tiny apps on their own would also benefit from memchr since they're often byte-centric, and those are no-std because of using tiny-std.

osiewicz:
To add to the list of potential use cases, I have a Windows kernel driver with `std` disabled. Nice PR @MarcusGrass !

BurntSushi (approved):
Nice, thanks!

Issue #124, opened 2023-05-29
failed to compile `memchr 2.5.0` on macos-12 and macos-11


chenrui333:
While trying to [upgrade gifski to 1.11.0](https://github.com/Homebrew/homebrew-core/pull/129833), we found some compilation error with `memchr 2.5.0` on macos-12 and macos-11

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: NotPresent', /Users/caleb/Library/Caches/Homebrew/cargo_cache/registry/src/index.crates.io-6f17d22bba15001f/memchr-2.5.0/build.rs:21:50
stack backtrace:
   0:        0x10002f1d6 - __mh_execute_header
   1:        0x10003fb9a - __mh_execute_header
   2:        0x10002193c - __mh_execute_header
   3:        0x10002efba - __mh_execute_header
   4:        0x10001d0a3 - __mh_execute_header
   5:        0x10001cd92 - __mh_execute_header
   6:        0x10001dc15 - __mh_execute_header
   7:        0x10002f4f4 - __mh_execute_header
   8:        0x10002f2d9 - __mh_execute_header
   9:        0x10001d6dd - __mh_execute_header
  10:        0x100045223 - __mh_execute_header
  11:        0x1000455d5 - __mh_execute_header
  12:        0x100005f6d - __mh_execute_header
  13:        0x100007700 - __mh_execute_header
  14:        0x100007699 - __mh_execute_header
  15:        0x10000592a - __mh_execute_header
  16:        0x10000602d - __mh_execute_header
  17:        0x100005d40 - __mh_execute_header
  18:        0x10001c3a4 - __mh_execute_header
  19:        0x100005d18 - __mh_execute_header
  20:        0x100007f48 - __mh_execute_header
Process 58184 exited with status = 101 (0x00000065)
```

full build log, https://github.com/Homebrew/homebrew-core/actions/runs/5083561622/jobs/9134939048
relates to https://github.com/Homebrew/homebrew-core/pull/129833

BurntSushi:
So #125 passes, which to me means this isn't a `macOS-{11,12}` issue?

The error seems to be coming from this line: https://github.com/BurntSushi/memchr/blob/be564d425bfcabd2c9925a8a360f4a62b866bb90/build.rs#LL21C26-L21C47

[...]

PR #125, opened 2023-05-29
ci: include macos-11,13

chenrui333:
include more osx runners for CI

relates to https://github.com/BurntSushi/memchr/issues/124


BurntSushi:
CI passes here so it doesn't seem to help. Therefore I'm going to close this.

See also: https://github.com/BurntSushi/memchr/issues/124#issuecomment-1575558129

PR #127, opened 2023-07-11
crate: add 'alloc' feature

### LINKED ISSUE #121: Feature request: no-std + alloc
I was looking through the code of this crate. I have a need for something like this on a no-std + alloc target, but it seems several features (such as using `Cow ` from alloc) are missing. That should be possible to support.

BurntSushi:
This crate currently only has a 'std' feature and of course support for no-std. But in that mode, it was also no-alloc. In practice, this crate is virtually all core-only, that is, no-std and no-alloc. There are really only a few parts that want std or alloc. For std, runtime CPU feature detection requires it. For alloc, there are a few minor APIs that require allocation.

Previously, the only way to get the APIs that require allocation was to enable 'std'. This new feature permits getting those APIs without bringing in 'std'.

I don't expect this to unlock any new or interesting use cases, but it will make downstream configurations a little more sensible and less annoying.

Fixes #121

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,11 @@ jobs:
       name: Show byte order for debugging
       run: ${{ env.CARGO }} test --verbose $TARGET byte_order -- --nocapture
     - if: matrix.build != 'pinned'
+      name: Run tests under default configuration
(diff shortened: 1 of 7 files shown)


BurntSushi:
I'm also rolling #120 and #123 into this PR.

PR #132, opened 2023-08-29
use `alloc` instead of `std` for `FinderRev::into_owned`

BurntSushi:
This was an oversight in the 2.6 release. The forward case was marked as `alloc` (which was a new feature as of 2.6) but the reverse case was not.


BurntSushi:
This is on crates.io in `memchr 2.6.1`.

PR #134, opened 2023-08-30
unsafe impl Sync+Send

### LINKED ISSUE #133: `Memchr<'_>` is no longer `Sync` from v2.6
Test code:

```rust
use memchr::memchr_iter;

pub fn assert_sync<T: Sync>(_: &T) {}

pub fn foo(input: &[u8]) {
    assert_sync(&memchr_iter(b'\n', input));
}
```

See also https://github.com/Nugine/s3s/issues/79

conradludgate:
Fixes #133 

The pointers in `Iter` seem to be simply an optimization over using `&'h [u8]` which is Send+Sync. I couldn't find a more straightforward solution than just implementing Send+Sync on Iter.

If my safety comments aren't sufficient, I can improve them

### DIFF
--- a/src/arch/generic/memchr.rs
+++ b/src/arch/generic/memchr.rs
@@ -1011,6 +1011,11 @@ pub(crate) struct Iter<'h> {
     haystack: core::marker::PhantomData<&'h [u8]>,
 }
 
+// SAFETY: `Iter` contains no shared references to anything that performs any interior mutations.
+unsafe impl Send for Iter<'_> {}
+// SAFETY: `Iter` perform no interior mutations, therefore no explicit synchronisation is necessary.
+unsafe impl Sync for Iter<'_> {}
+
 impl<'h> Iter<'h> {
     /// Create a new generic memchr iterator.
     #[inline(always)]
--- a/src/tests/memchr/mod.rs
+++ b/src/tests/memchr/mod.rs
@@ -305,3 +305,11 @@ impl Seed {
         more.into_iter()
     }
 }
+
+fn sync_regression() {
+    use crate::{Memchr, Memchr2, Memchr3};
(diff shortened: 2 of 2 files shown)


src/arch/generic/memchr.rs
```diff
@@ -1011,6 +1011,11 @@ pub(crate) struct Iter<'h> {
     haystack: core::marker::PhantomData<&'h [u8]>,
 }
 
+// SAFETY: `Iter` contains no shared references to anything that performs any interior mutations.
```
BurntSushi, line 1014: Code/comments should respect the surrounding style and be wrapped to 79 columns (inclusive). Similarly below.

src/tests/memchr/mod.rs
```diff
@@ -305,3 +305,11 @@ impl Seed {
         more.into_iter()
     }
 }
+
+fn sync_regression() {
+    use crate::{Memchr, Memchr2, Memchr3};
+    fn assert_send_sync<T: Send + Send>() {}
```
BurntSushi, line 311: This should be `Send + Sync`. And also include `UnwindSafe` and `RefUnwindSafe`, as those are two other auto traits I've accidentally dropped in the past in other crates.

Also should be moved to the test section of `src/memchr.rs` and marked with `#[test]`.

BurntSushi (changes requested):
Lovely, thank you! I've fixed up the PR (see feedback) and merged it in https://github.com/BurntSushi/memchr/commit/a12d405fa7225d42bb125e9f2178516aaf4104ed.

src/tests/memchr/mod.rs
conradludgate: > This should be Send + Sync

Ah, oops!

PR #143, opened 2023-12-28
Remove obsolete exclusions

atouchet:
I no longer see any bench or tmp directories in the repo.


BurntSushi:
Note that I left `/tmp`. It may not be in the repo, but it sometimes exists.

PR #149, opened 2024-03-27
require 'simd128' on 'wasm32'

### LINKED ISSUE #144: The force-enabling of SIMD128 feature can lead to unloadable WASMs in some browsers
In several cases, memchr force-enables the SIMD feature on wasm32 (for example [here](https://github.com/BurntSushi/memchr/blob/cedf318090876c6d557f234b158dd4fdc91c41ec/src/arch/wasm32/simd128/packedpair.rs#L63)). This has been an issue for me because, in a fairly large project where memchr is used indirectly by several packages (nom, nom_locate, pulldown-cmark and regex), the code with SIMD gets pulled in and the resulting WASM fails to load on some slightly-older browsers such as Safari iOS 15, with an error looking like:
```
CompileError: WebAssembly.Module doesn't parse at byte X: can't get Function local's type in group Y, in function at index Z
```

An example of code pulled in (as shown by `wasm-decompile`) is:
```
function memchr_arch_wasm32_simd128_packedpair_Finder_with_pair_impl_hd8b88a6eee7061af(a:long_ptr, b:int, c:int, d:int, e:int)
```

In release mode the code has so far been optimized out but in dev mode it is still in, although I did not specify `target-feature=+simd128` in my `RUSTFLAGS`, so it is a problem to test/debug with older Safari. I'm also worried whether in some cases such code might find its way to the release mode as well.

A possible work-around would be to add more aggressive dead code elimination in dev mode but I did not find how to do that.

BurntSushi:
It turns out that providing routines with `#[target_feature(enable =
"simd128")]` on `wasm32` can fail in some older browsers. The specific
problem is not totally clear to me, but it is straight-forward enough
to fix (I hope) by just requiring that `simd128` be enabled at compile
time in order to include the `wasm32` SIMD modules in this crate.

This would not be a great solution if WASM supported runtime CPU
feature detection. And the status quo is that `simd128` has to be
enabled at compile time anyway for the SIMD code to take effect. So
this shouldn't cause any regressions and is likely something we can do
long term as well. We can re-evaluate once and if WASM gets support for
runtime CPU feature detection.

We also add a CI test for `wasm32` *without* the `simd128` target
feature enabled.

Fixes #144

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
           target: s390x-unknown-linux-gnu
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
@@ -138,7 +138,7 @@ jobs:
       WASMTIME_VERSION: 15.0.1
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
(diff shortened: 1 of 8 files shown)


BurntSushi:
This PR is in `memchr 2.7.2` on crates.io.

PR #158, opened 2024-07-29
Fix clippy warnings, add clippy to CI

purplesyringa:
This should hopefully ease maintenance of the crate.


BurntSushi:
The transition to `wasip1` LGTM.

But I don't have Clippy enabled for a reason. There are a lot of lints that I don't always agree with. Even in this PR, I see some changes I like and some I don't. I'm not sure I really want to spend the time going through all of them to be honest.

I have a long-standing TODO item to go through the Clippy lints, select the ones I like and then use them in my various Rust projects. But it hasn't been a priority for me.

purplesyringa:
How about I apply the least contentious clippy lints for now? I'd like to:
- Fix spaces in doc comments
- Replace the use of `needles.get(0)` / `needles.get(1)` / etc. with the `match` block
- Replace `core::u*::MAX` with `u*::MAX`
This alone would remove a significant percentage of clippy lints.

BurntSushi:
I agree with the third bullet point, but not the first two. For (1), my line wrapper doesn't conform to the style that Clippy wants, so that would be annoying for me. For (2), I don't have a super strong opinion, but I actually like the status quo better.

purplesyringa:
I see. I think I'll apply the 3rd bullet then, and whitelist other lints. That should be no worse than status quo and might actually catch bugs.

BurntSushi:
Please don't include Clippy in CI.

purplesyringa:
Sure, I'll get to it soon. What is the rationale by the way?

purplesyringa:
I have reduced the PR to just wasm transition, the core:: replacement, and a bunch of clippy allows. Is that sufficient?

BurntSushi:
> Sure, I'll get to it soon. What is the rationale by the way?

I think I explained why here: https://github.com/BurntSushi/memchr/pull/158#issuecomment-2256645758

> I have reduced the PR to just wasm transition, the core:: replacement, and a bunch of clippy allows. Is that sufficient?

No. Please remove Clippy from CI. I've already said that I have a TODO to go through Clippy lints and pick the ones I want. But I haven't done it yet, and until then, I don't want Clippy in CI.

purplesyringa:
Oh, wait, I must have misunderstood you. I assumed you don't want clippy *in CI* but still want it outside CI. That was odd phrasing. Fixed.

BurntSushi (approved):
Thanks!

purplesyringa:
I'm wondering if you still wish to merge this?

PR #162, opened 2024-09-29
Make this crate build on big-endian aarch64.

he32:
As noted in https://github.com/rust-lang/stdarch/issues/1484, the NEON intrinsics are broken on big-endian aarch64.

This is part of fixing rust to build for & on big-endian aarch64, following up https://github.com/rust-lang/rust/issues/129819.

### DIFF
--- a/src/arch/aarch64/memchr.rs
+++ b/src/arch/aarch64/memchr.rs
@@ -8,7 +8,7 @@ available for `aarch64` targets.)
 
 macro_rules! defraw {
     ($ty:ident, $find:ident, $start:ident, $end:ident, $($needles:ident),+) => {{
-        #[cfg(target_feature = "neon")]
+        #[cfg(all(target_feature = "neon", target_endian = "little"))]
         {
             use crate::arch::aarch64::neon::memchr::$ty;
 
@@ -19,7 +19,7 @@ macro_rules! defraw {
             // enabled.
             $ty::new_unchecked($($needles),+).$find($start, $end)
         }
-        #[cfg(not(target_feature = "neon"))]
+        #[cfg(not(all(target_feature = "neon", target_endian = "little")))]
         {
             use crate::arch::all::memchr::$ty;
 
--- a/src/arch/aarch64/mod.rs
+++ b/src/arch/aarch64/mod.rs
@@ -2,6 +2,8 @@
 Vector algorithms for the `aarch64` target.
 */
 
+#[cfg(target_endian = "little")]
 pub mod neon;
 
+#[cfg(target_endian = "little")]
 pub(crate) mod memchr;
--- a/src/memchr.rs
+++ b/src/memchr.rs
@@ -518,14 +518,14 @@ unsafe fn memchr_raw(
     {
         crate::arch::wasm32::memchr::memchr_raw(needle, start, end)
     }
-    #[cfg(target_arch = "aarch64")]
+    #[cfg(all(target_arch = "aarch64", target_endian = "little"))]
     {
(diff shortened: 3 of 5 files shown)


BurntSushi:
Can you add a big endian test target to CI?

he32:
> Can you add a big endian test target to CI?

Sorry, I have no idea how I do that.
I do have a local qemu-run big-endian NetBSD/aarch64 10.0 installation running.  However, it appears that for this to work, it depends on which version of qemu is used -- I'm currently using qemu 8.1.3, and there it works, but it does not look like qemu 8.0.2 works.  I do not think I've done testing for qemu 9.0.1.  Then there's the issue of how to match this up with CI, which I've never done before.

BurntSushi:
Have you looked into using cross? CI is already using cross for many targets (which uses qemu under the hood). Ideally you would find a big endian target and just add it to CI.

he32:
> Have you looked into using cross? CI is already using cross for
> many targets (which uses qemu under the hood). Ideally you
> would find a big endian target and just add it to CI.

"cross", what is that?

If you want to put a label on me, you can call me a pre-cloud
programmer, sysadmin and network administrator, so please have
some patience and provide explicit references.

BurntSushi:
I don't have a ton of time for mentoring, but if you're looking to fix big endian targets, then we should add testing for those targets to CI. Otherwise there is nothing ensuring that your fixes are correct or that the problems stay fixed. 

Please look at the existing CI configuration in `.github`. You should see cross being used.

BurntSushi:
See also https://github.com/cross-rs/cross

Cross has nothing to do with "the cloud."

he32:
Hmm...  This crate is vendored into the rust compiler.  WIthout these fixes, there will not be a big-endian aarch64 rust compiler without having to manually apply patches, and without the rust compiler, I predict there will be problems setting up the CI.  So how do we break this logjam?

BurntSushi:
What have you tried? What happened? Did you investigate cross?

he32:
> What have you tried? What happened? Did you investigate cross?

What brought me here was the appearance of a need to patch your
crate which is vendored into the rust 1.80.1 compiler tarball
source distribution, and my desire to make this build both for
and on NetBSD/aarch64eb.  I find it unlikely that the rust 1.80.1
compiler will build for any other big-endian aarch64 targets
without a similar patch.

The goal of my submitting this pull request is that the fix from
the pull request will eventually make it into the vendored crate
in the rust compiler, so that a future version will build for
big-endian aarch64 without having to apply and maintain these
fixes as patches.

I have therefore not yet studied enabling an aarch64eb CI target.

I highly suspects it's a "chicken and egg" situation.

E.g. there is no big-endian aarch64 target listed in
https://forge.rust-lang.org/infra/other-installation-methods.html

The point of this whole excercise is to get out of that
situation.

BurntSushi:
Fair enough. This means I'm going to need to investigate myself and I don't know when I'm going to have time to do that.

he32:
For what it's worth, I present to you the build result of your crate with my fixes applied,
run on a big-endian aarch64 host, ref. the attachment.
[typescript.txt](https://github.com/user-attachments/files/17235159/typescript.txt)

workingjubilee:
Please fix this by fixing the intrinsics upstream. Fixing this in every single crate is not less effort than fixing it in stdarch: https://github.com/rust-lang/stdarch/issues/1484#issuecomment-1942934003

Dr-Emann:
Presumably this is fixed, since #188 happened, the crate must build on big endian aarch64?

Issue #160, opened 2024-09-10
Feature Switch to Full Compile-Time Feature Detection


Sewer56:
Currently `memchr` does run-time feature detection for x86, storing 'preferred' functions inside a field that is later referenced in subsequent calls

```rust
// Inside `unsafe_ifunc!`

// store
FN.store(fun as Fn, Ordering::Relaxed);

// use
let fun = FN.load(Ordering::Relaxed);
```

This manifests itself as a `mov` and `call` in x86; and one or two other misc instructions along the way.

In order to avoid unnecessary code bloat and improve performance, a feature to bypass this mechanism could be introduced.
This would allow for inlining, and slightly lesser call overhead.

That is of course, when the Rust compiler is invoked with the correct flags to ship a binary targeting a specific processor.

Note
------

This isn't a formal feature request, it's more so a reminder for myself.
I want to reference this issue in my own codebase, and possibly PR this sometime in the future.

BurntSushi:
I think if would be better to try and do this without feature flags. We can detect at compile time whether, for example, the avx2 target feature is enabled.

Sewer56:
> I think if would be better to try and do this without feature flags.

I wrote this when I was tired (still am).
Seems you're right.

Ambyjkl:
I haven't examined the code here yet, but it seems just setting rustflags to `-Ctarget-feature=+aes2` might be enough to skip the check, not sure https://github.com/BurntSushi/memchr/pull/127/commits/b4a4c6aabd796f768b68c89a5f5eb6f18f44d4a2

Sewer56:
https://godbolt.org/z/h7av5rx3n

```rust
extern crate memchr;
use memchr::memchr;

#[no_mangle]
fn memchr_1(data: &[u8]) -> usize {
    memchr(0, &data).unwrap()
}
```

![image](https://github.com/user-attachments/assets/c57fb1a9-2c3f-491d-8a80-6840e66cda30)

I believe these `mov` operation correspond to

```rust
let fun = FN.load(Ordering::Relaxed);
```

Ambyjkl:
I see, seems the compiler doesn't automatically inline the call, sad

Ambyjkl:
[...]

PR #165, opened 2024-10-09
Improve: Call target directly from unsafe_ifunc macro whenever possible.

Sewer56:
Example based on discussion from:

- https://github.com/BurntSushi/memchr/issues/160


Sewer56:
CI Failure is not caused by PR.

This Actions Run
https://github.com/BurntSushi/memchr/actions/runs/11258409805

Was ran with commit reverted, i.e. using code from `master`.

PR #169, opened 2024-12-01
WIP: add support for AVX512

marxin:
First of all: I would like to thank you for the library!

I was just curious how difficult it would be to add support for AVX 512 vector types,  thus I made a proof-of-concept PR. The biggest complication for AVX512 comes from the fact the [cmp_eq function](https://doc.rust-lang.org/beta/core/arch/x86/fn._mm512_cmp_epi8_mask.html) has `u64` as a return type. That's a difference from `__m128i` and `__m256i` types, where the return type is a mask in the corresponding victory type. That's why the change would need a split of `Vector` trait into 2: `Vector` and `ComparisonVector`.

Anyway, the question is if you'll be interested in the full support?

### DIFF
--- a/src/arch/generic/memchr.rs
+++ b/src/arch/generic/memchr.rs
@@ -92,7 +92,7 @@ Generic crate-internal routines for the `memchr` family of functions.
 
 use crate::{
     ext::Pointer,
-    vector::{MoveMask, Vector},
+    vector::{ComparisonVector, MoveMask, Vector},
 };
 
 /// Finds all occurrences of a single byte in a haystack.
@@ -149,7 +149,7 @@ impl<V: Vector> One<V> {
         // need to move up to using a u64 for the masks used below. Currently
         // they are 32 bits, which means we're SOL for vectors that need masks
         // bigger than 32 bits. Overall unclear until there's a use case.
-        debug_assert!(V::BYTES <= 32, "vector cannot be bigger than 32 bytes");
+        debug_assert!(V::BYTES <= 64, "vector cannot be bigger than 64 bytes");
 
         let topos = V::Mask::first_offset;
         let len = end.distance(start);
--- a/src/arch/generic/packedpair.rs
+++ b/src/arch/generic/packedpair.rs
@@ -11,7 +11,7 @@ frequencies to heuristically select the pair of bytes to search for.
 use crate::{
     arch::all::{is_equal_raw, packedpair::Pair},
     ext::Pointer,
-    vector::{MoveMask, Vector},
+    vector::{ComparisonVector, MoveMask, Vector},
 };
 
 /// A generic architecture dependent "packed pair" finder.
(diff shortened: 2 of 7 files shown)


BurntSushi:
I'm definitely not opposed to it in principle, but...

* As I understand it, AVX-512 is being phased out of consumer chips. For example, my i9-12900K has zero AVX-512 support.
* Using AVX-512 can result in down-clocking. I think this was true of AVX2 at one point, but my understanding was that this was more severe for AVX-512.
* The extent to which I'm willing to support AVX-512 depends on just how much it improves performance. That is, it would be nice to see some benchmarks.
* Perhaps most importantly, AVX-512 support is unstable. And my general policy is not to support unstable things. I've done this for a variety of things in the past and it has really been nothing but a headache. And this cost is largely bared by me.

marxin:
> * As I understand it, AVX-512 is being phased out of consumer chips. For example, my i9-12900K has zero AVX-512 support.

Yep, I am aware of the limitation. However, it seems the modern AMD Zen CPUs have the capability: https://en.wikipedia.org/wiki/AVX-512#CPUs_with_AVX-512.

> * Using AVX-512 can result in down-clocking. I think this was true of AVX2 at one point, but my understanding was that this was more severe for AVX-512.

That's not the case for AMD CPUs if I remember correctly, but yes, it's an unpleasant limitation that some of the CPUs have the limitation.

> * The extent to which I'm willing to support AVX-512 depends on just how much it improves performance. That is, it would be nice to see some benchmarks.

I decided to make a micro-benchmark that runs `count` version of the memchr function for various lengths:
https://github.com/marxin/memchr-avx2-avx512-benchmark

I was able to get the following numbers on my AMD Ryzen 9 7900X 12-Core Processor (4th gen Zen CPU):
```
Size        AVX512/AVX2 (smaller is better)
    4 KiB:     0.778
   16 KiB:     0.705
   32 KiB:     0.68
   64 KiB:     0.676
  128 KiB:     0.665
  256 KiB:     0.659
  512 KiB:     0.661
 1024 KiB:     0.811
 2048 KiB:     0.729
 4096 KiB:     0.78
 8192 KiB:     0.817
16384 KiB:     0.825
32768 KiB:     0.855
65536 KiB:     0.847
```

So it seems, one can get about 20% speed-up for smaller sizes, and about 10-15% for much bigger. Compared to the other ISA extensions, are these good numbers :P ?

> * Perhaps most importantly, AVX-512 support is unstable. And my general policy is not to support unstable things. I've done this for a variety of things in the past and it has really been nothing but a headache. And this cost is largely bared by me.

Yeah, I've got that. Then, we can probably return to it once it's stable.

BurntSushi:
> Then, we can probably return to it once it's stable.

Sounds good.

marxin:
Thanks for pointing to the numbers.

> And remember that these benchmarks aren't the only consideration due to, AIUI, AVX-512 down-clocking the CPU in at least some cases.
> 
> Generally speaking, I find "not true for AMD CPUs" to be uncompelling overall. It is possible that if we were to add AVX-512 support, it would need to be opt-in. Whether via a Cargo feature or via the API, I'm not sure.

Yeah, I fully agree the usage of AVX-512 during run-time is problematic due to the CPU down-clocking on many existing pieces of hardware. It's not a pure win as once AVX2 compared to SSEx :)

PR #171, opened 2024-12-31
loongarch64: add vector routines

heiher:
This patch adds `LSX` vector routines for LoongArch64, significantly improving performance. See the commit message for details.

### DIFF
--- a/src/arch/loongarch64/lsx/memchr.rs
+++ b/src/arch/loongarch64/lsx/memchr.rs
@@ -1,0 +1,1031 @@
+/*!
+This module defines 128-bit vector implementations of `memchr` and friends.
+
+The main types in this module are [`One`], [`Two`] and [`Three`]. They are for
+searching for one, two or three distinct bytes, respectively, in a haystack.
+Each type also has corresponding double ended iterators. These searchers are
+typically much faster than scalar routines accomplishing the same task.
+
+The `One` searcher also provides a [`One::count`] routine for efficiently
+counting the number of times a single byte occurs in a haystack. This is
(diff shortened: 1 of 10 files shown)


BurntSushi:
Thanks! This looks like a fair bit of work.

But I really wish folks would file issues before throwing up huge PRs like this. Generally speaking, I don't let nightly code into my crates without a really compelling motivation. I've tried it before, and it just basically always results in headaches in one way or another. And this is especially fraught because I don't have access to any LoongArch hardware. Which brings me to my next major problem with this PR: this doesn't test any of these changes in CI. Can we run LoongArch tests somehow? If not, that seems like a major problem that needs to be addressed. I see that it is [listed as a supported target for Cross](https://github.com/cross-rs/cross), so hopefully that's easy to address.

A more minor problem is that because I don't have any LoongArch hardware, I can't actually run the benchmarks. I don't think that can be meaningfully solved by Cross. I don't think this is a blocking issue ultimately.

heiher:
Thank you for your comments. I'll add CI tests for the LoongArch targets in a new PR. This PR will be updated after the stabilization of the dependent SIMD target features. As for performance benchmarking, I propose that I track it using the physical machines in my CI (due to network instability, I'm currently unable to access GitHub).

BurntSushi:
CI tests should be in this PR, please.

My suggestion is that we close this PR for now and we can revisit this once the intrinsics are stabilized.

heiher:
> CI tests should come with the new target support, please.
> 
> My suggestion is that we close this PR for now and we can revisit this once the intrinsics are stabilized.

Okay.

PR #172, opened 2025-01-03
ci: add support for loongarch64 targets

heiher:
The latest release of `cross` does not support the LoongArch target yet. This PR switches to installing a pinned commit from Git.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,8 @@ jobs:
       # Bump this as appropriate. We pin to a version to make sure CI
       # continues to work as cross releases in the past have broken things
       # in subtle ways.
-      CROSS_VERSION: v0.2.5
+      CROSS_GIT_URL: https://github.com/cross-rs/cross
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -122,13 +122,13 @@ jobs:
     - name: Show byte order for debugging
       run: ${{ env.CARGO }} test --verbose $TARGET byte_order -- --nocapture
     - name: Run tests
-      run: ${{ env.CARGO }} test --verbose
+      run: ${{ env.CARGO }} test --verbose $TARGET
```
heiher, line 125: Was this missed before this?

BurntSushi: Possibly, It's just a diagnostic and it's rarely anything other than little endian, so definitely easy to miss.

BurntSushi (changes requested):
I really don't want to get into the business of pinning to specific `cross` commits. And I don't want to add a target to CI if I can't rely on it working. Maybe Rust support for `loongarch64` needs to mature a bit first.

PR #180, opened 2025-06-06
Remove `compiler-builtins` from `rustc-dep-of-std` dependencies

tgross35:
Since [1], this will come automatically from `rustc-std-workspace-core` and the crates.io dependency should no longer be specified.

[1]: https://github.com/rust-lang/rust/pull/141993


BurntSushi:
Is there any risk of removing this causing `memchr` to stop working with older versions of Rust? (I don't think so, but I wanted to ask.)

tgross35:
Awesome, thank you!

PR #182, opened 2025-06-26
memmem: add owned finder constructor variants

### LINKED ISSUE #173: Directly build an owned `Finder`
Hi :)

I have a use case where I already have an owned needle in the form of `Vec<u8>` / `Box<[u8]>` and we would like to build an owned `Finder`.
As far as I aware, the way to go is to:
```Rust
let finder = Finder:new(&bytes).into_owned();
```

but this will pass the bytes by reference, just to heap allocate them again in the `into_owned()` call.
Is there a way to directly build an owned `Finder` without requiring an additional heap allocation and drop? I am very sorry if I have missed it!
The code in question is very performance sensitive and we would like to avoid paying any extra cost that we think could be avoided.

If you think this is something that this crate could offer, I am happy to work on the implementation.

Thank you in advance!

marmeladema:
I wasn't sure how to call the new constructors so fell back to just suffixing already existing ones with `_owned`.

Note that I haven't added the `From<Box<[u8]>>` (and related) convenient helpers since they are not strictly required and I would like to land the minimal required change first.

Fixes #173

### DIFF
--- a/src/cow.rs
+++ b/src/cow.rs
@@ -43,7 +43,9 @@ impl<'a> CowBytes<'a> {
     /// Create a new owned CowBytes.
     #[cfg(feature = "alloc")]
     #[inline(always)]
-    fn new_owned(bytes: alloc::boxed::Box<[u8]>) -> CowBytes<'static> {
+    pub(crate) fn new_owned(
+        bytes: alloc::boxed::Box<[u8]>,
+    ) -> CowBytes<'static> {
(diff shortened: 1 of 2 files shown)


marmeladema:
Gentle ping @BurntSushi :) I would like to use this in https://github.com/cloudflare/wirefilter but i'd need this improvement to land first 👍

BurntSushi:
OK, I've updated the PR to remove the `?Sized` bound.

PR #185, opened 2025-07-20
arch::generic::memchr::OneMatches iterator proof-of-concept

Yomguithereal:
Hello @BurntSushi, here is a dummy PR meant as a conversation opener regarding the inclusion of an iterator for `memchr` routines with different amortization principles (as previously discussed in #184). The idea is to have an iterator for cases when people actually know they will need to scan the whole haystack for matches and want to amortize the vectorized calls along the way because some matches can sometimes be very close to one another.

Several notes:

* I am not sure the boolean tracking whether the first unaligned load was performed is actually required. I think it can be inferred from a `self.start == self.current` test, but I am not sure.
* The trick about the last unaligned load is not possible in this iterator's context since its invariant does not hold, since we may stumble upon a match reported earlier (we could store an offset, though, but is it worth the cost?).
* The iterator does not perform any unrolling because it would probably mean storing more than one mask at a time.

Please feel free to tell me to let the matter rest if you want, I don't want to be a bother.

### DIFF
--- a/src/arch/generic/memchr.rs
+++ b/src/arch/generic/memchr.rs
@@ -428,6 +428,108 @@ impl<V: Vector> One<V> {
     }
 }
 
+pub(crate) struct OneMatches<'a, 'h, V: Vector> {
+    /// The starting point into the haystack. We use this to convert
+    /// pointers to offsets.
+    start: *const u8,
+    /// The ending point into the haystack.
+    end: *const u8,
+    /// The current position into the haystack.
+    current: *const u8,
+    /// Whether we have read first unaligned chunk.
+    has_read_initial_unaligned_chunk: bool,
+    /// The current move mask being processed.
+    mask: Option<V::Mask>,
+    /// Reference to searcher.
+    searcher: &'a One<V>,
+    /// A marker for tracking the lifetime of the start/end/current
+    /// pointers above, which all point into the haystack.
+    haystack: core::marker::PhantomData<&'h [u8]>,
+}
+
+impl<'a, 'h, V: Vector> OneMatches<'a, 'h, V> {
+    fn new(searcher: &'a One<V>, haystack: &'h [u8]) -> Self {
+        Self {
+            start: haystack.as_ptr(),
+            end: haystack.as_ptr().wrapping_add(haystack.len()),
+            current: haystack.as_ptr(),
+            has_read_initial_unaligned_chunk: false,
+            mask: None,
+            searcher,
+            haystack: core::marker::PhantomData,
+        }
+    }
+
+    unsafe fn next(&mut self) -> Option<usize> {
+        'main: loop {
+            // Processing current move mask
+            if let Some(mask) = &mut self.mask {
+                debug_assert!(mask.has_non_zero());
+
+                let offset = self.current.add(mask.first_offset());
+                let next_mask = mask.clear_least_significant_bit();
+
+                if next_mask.has_non_zero() {
+                    *mask = next_mask;
+                } else {
+                    self.mask = None;
+                }
+
+                return Some(offset.distance(self.start));
+            }
+
+            // Initial unaligned load
+            if self.has_read_initial_unaligned_chunk {
+                self.has_read_initial_unaligned_chunk = true;
+
+                let chunk = V::load_unaligned(self.current);
+                let mask = self.searcher.v1.cmpeq(chunk).movemask();
+
+                self.current = self
+                    .current
+                    .add(V::BYTES - (self.start.as_usize() & V::ALIGN));
+
+                if mask.has_non_zero() {
+                    self.mask = Some(mask);
+                    continue 'main;
+                }
+            }
+
+            // Main loop of aligned loads
+            while self.current <= self.end.sub(V::BYTES) {
+                debug_assert_eq!(0, self.current.as_usize() % V::BYTES);
+
+                let chunk = V::load_aligned(self.current);
+                let mask = self.searcher.v1.cmpeq(chunk).movemask();
+
+                self.current = self.current.add(V::BYTES);
+
+                if mask.has_non_zero() {
+                    self.mask = Some(mask);
+                    continue 'main;
+                }
+            }
+
+            // Processing remaining bytes linearly
+            if self.current < self.end {
+                if let Some(offset) =
+                    fwd_byte_by_byte(self.current, self.end, |b| {
+                        b == self.searcher.needle1()
+                    })
+                {
+                    // At most, adding one to offset here should reach `end`.
+                    self.current = offset.add(1);
+                    return Some(offset.distance(self.start));
+                }
+
+                self.current = self.end;
+            }
+
+            return None;
+        }
+    }
+}
+
 /// Finds all occurrences of two bytes in a haystack.
 ///
 /// That is, this reports matches of one of two possible bytes. For example,


BurntSushi:
I think before I'm willing to expand the API and make it more complicated, there needs to be more benchmarking:

1. The status quo
2. Changing the current iterator to support this sort of amortization
3. Providing both options

This would give us concrete numbers to weigh against the API complexity of more iterators.

Yomguithereal:
Hello @BurntSushi, I have started an early benchmark over there to address the question: https://github.com/Yomguithereal/memchr-iter-bench

I am new to the "benchmarking in Rust" game so I used `criterion`, I hope this is a sensible choice. The results (the formatting is a bit silly sorry, the only thing I found to quickly render the results as a table is `criterion-table`) are [here](https://github.com/Yomguithereal/memchr-iter-bench/blob/master/RESULTS.md).

I benchmarked a naive loop of `memchr`, an amortized loop of `memchr` (the splat is only created once), `memchr_iter` and the proposed (fixed, lol the PR code is wildly incorrect since I was unable to test it) iterator (I will refer to it as `memchr_memoized` later on), all against a linear scalar baseline. I benchmarked all those options on strings ranging from short to long, also varying the density of matches found within it.

I only benchmarked `sse2` instructions currently because I felt they would yield the least performance vs. other SIMD implementations and the 16 bytes step should hurt `memchr_memoized` more, in principle.

I only benchmarked the case where you are searching for a single needle yet.

Also, the use case here is to report all matches, not only the first one.

Some things I seem to observe:

* Amortizing a single splat vector creation seems to be pointless
* `memchr_iter` is usually a tiny bit faster than the naive or amortized `memchr` loop, as you hinted at
* `memchr_memoized` is usually competitive when strings are long and when they are expected to contain multiple needles, the denser the better.
* On the degenerate case where the haystack only contains needles or nearly only needles, `memchr_memoized` does a better job than `memchr` etc, of course, but is still worse than the scalar solution, which makes sense

It should also be noted that, for now, `memchr_memoized` does not try to load aligned pointer at all, notably because the iterator logic and the fact you need to report all matches make the clever unaligned tricks awkward since you need to avoid overlapping which would mean reporting multiple times the same match. There are of course ways around that, in order to load aligned pointers but I have not worked toward this yet.

So I think `memchr_memoized` has good value for a specific use-case but still represent some kind of trade-off based on what you need to do, which means it is probably not a good fit for a replacement of `memchr_iter`, right?

BurntSushi:
Nice work. Note that what you have isn't a one-shot searcher. It is prebuilt. One-shot means the search is completely reconstructed on every call.

Your benchmark results look nice. The biggest wins seem to come from benchmarks with lots of matches, which is exactly what I'd expect.

Anyway, this is going to require a deeper review from me. And there's still a fair bit of work needed for a proper comparison here. You've hand-coded something _outside_ the `memchr` crate, but there's infrastructure that's missing (particularly around CPU feature detection). Ideally what the comparison would look like is the status quo versus a `memchr` crate whose memchr iterators work like yours. i.e., No API differences. That would be the best apples-to-apples comparison.

Yomguithereal:
> Nice work. Note that what you have isn't a one-shot searcher. It is prebuilt. One-shot means the search is completely reconstructed on every call.

You are right indeed. I have renamed the engine accordingly.

I will try to benchmark AVX2 now and see whether some strategy aligning pointers yields better results also.

> Anyway, this is going to require a deeper review from me. And there's still a fair bit of work needed for a proper comparison here. You've hand-coded something outside the memchr crate, but there's infrastructure that's missing (particularly around CPU feature detection). Ideally what the comparison would look like is the status quo versus a memchr crate whose memchr iterators work like yours. i.e., No API differences. That would be the best apples-to-apples comparison.

I can try to do so if I can find enough time.

BurntSushi:
I don't. I'm not really an ISA extension expert. I've generally always seen an improvement when using AVX2 compared to SSE2.

I do know that on some hardware AVX2 can result in down-clocking the CPU, but I thought that was on very old CPUs. But it might be worth trying other hardware.

My understanding is that, at least on x86-64, pointer alignment doesn't really matter for performance. I've never really been able to detect a measurable difference. But I could be wrong.

BurntSushi:
Oh that's interesting. I wish I had time to dig into this with you. I suspect there is something wrong. The fact that `memchr/sherlock/never/huge2` is twice as slow with bufchr suggests something is pretty wrong. That benchmark shouldn't rely at all on the iterator optimization since there's never any matches. If I were working on this, I'd look toward carefully scrutinizing the codegen.

Yomguithereal:
Ok, so the explanation seems to be that I am a buffoon that succeeded in improperly enabling the avx2 feature, prompting the compiler to fallback to unoptimized sse2 instructions in the generated code (I am not entirely sure how this is even possible, I would expect the compiler to not compile or produce a UB-producing executable instead of this, but I don't have enough skills in the matter to know what really happened here).

Now that I have avx2 instructions properly used, here is what I observe:

```txt
benchmark                          avx2 bufchr/memchr/prebuilt  avx2-aligned bufchr/memchr/prebuilt  memchr/memchr/prebuilt  sse2 bufchr/memchr/prebuilt
---------                          ---------------------------  -----------------------------------  ----------------------  ---------------------------
memchr/sherlock/common/huge1       4.2 GB/s (1.00x)             4.1 GB/s (1.02x)                     2.2 GB/s (1.90x)        3.4 GB/s (1.24x)
memchr/sherlock/common/small1      11.5 GB/s (1.04x)            8.4 GB/s (1.42x)                     2.4 GB/s (4.87x)        11.9 GB/s (1.00x)
memchr/sherlock/common/tiny1       2.8 GB/s (1.10x)             1827.9 MB/s (1.71x)                  901.4 MB/s (3.48x)      3.1 GB/s (1.00x)
memchr/sherlock/never/huge1        41.4 GB/s (2.84x)            45.2 GB/s (2.60x)                    117.6 GB/s (1.00x)      37.6 GB/s (3.13x)
memchr/sherlock/never/small1       19.9 GB/s (1.48x)            20.6 GB/s (1.43x)                    29.4 GB/s (1.00x)       15.9 GB/s (1.86x)
memchr/sherlock/never/tiny1        3.8 GB/s (1.00x)             2.6 GB/s (1.47x)                     3.6 GB/s (1.06x)        3.8 GB/s (1.00x)
memchr/sherlock/never/empty1       14.00ns (1.00x)              15.00ns (1.07x)                      16.00ns (1.14x)         14.00ns (1.00x)
memchr/sherlock/rare/huge1         37.1 GB/s (2.57x)            41.5 GB/s (2.30x)                    95.5 GB/s (1.00x)       34.1 GB/s (2.80x)
memchr/sherlock/rare/small1        21.3 GB/s (1.12x)            18.2 GB/s (1.31x)                    23.8 GB/s (1.00x)       17.7 GB/s (1.35x)
memchr/sherlock/rare/tiny1         3.8 GB/s (1.00x)             2.6 GB/s (1.47x)                     2.9 GB/s (1.29x)        3.8 GB/s (1.00x)
memchr/sherlock/uncommon/huge1     14.0 GB/s (1.00x)            11.6 GB/s (1.20x)                    7.4 GB/s (1.90x)        11.4 GB/s (1.23x)
memchr/sherlock/uncommon/small1    18.7 GB/s (1.00x)            14.7 GB/s (1.27x)                    9.5 GB/s (1.97x)        16.7 GB/s (1.12x)
memchr/sherlock/uncommon/tiny1     3.6 GB/s (1.00x)             2.5 GB/s (1.44x)                     1495.5 MB/s (2.44x)     3.6 GB/s (1.00x)
memchr/sherlock/verycommon/huge1   3.3 GB/s (1.00x)             2.6 GB/s (1.26x)                     1027.4 MB/s (3.32x)     2.5 GB/s (1.31x)
memchr/sherlock/verycommon/small1  8.0 GB/s (1.07x)             5.2 GB/s (1.64x)                     1064.3 MB/s (8.26x)     8.6 GB/s (1.00x)
```

Ranking:

```txt
Engine                               Version  Geometric mean of speed ratios  Benchmark count
------                               -------  ------------------------------  ---------------
avx2 bufchr/memchr/prebuilt          0.0.1    1.20                            15
sse2 bufchr/memchr/prebuilt          0.0.1    1.30                            15
avx2-aligned bufchr/memchr/prebuilt  0.0.1    1.46                            15
memchr/memchr/prebuilt               2.7.4    1.88                            15
```

So now avx2 is giving an edge, but not by much. Alignement seems to help a little bit with large inputs where the needle is rare, but not that much either. Also `memchr` is stil 2-3 times faster in this particular use-case, which is still a bit odd because I don't expect the loop unrolling to give such an edge?

When searching for 2 needles (using avx2 properly this time), the discrepancy drops:

```txt
benchmark                        rust/bufchr/memchr2/prebuilt  rust/memchr/memchr2
---------                        ----------------------------  -------------------
memchr/sherlock/common/huge2     3.1 GB/s (1.00x)              1337.4 MB/s (2.34x)
memchr/sherlock/common/small2    9.4 GB/s (1.00x)              1490.0 MB/s (6.44x)
memchr/sherlock/never/huge2      51.9 GB/s (1.27x)             65.9 GB/s (1.00x)
memchr/sherlock/never/small2     19.3 GB/s (1.10x)             21.3 GB/s (1.00x)
memchr/sherlock/never/tiny2      3.6 GB/s (1.00x)              3.6 GB/s (1.00x)
memchr/sherlock/never/empty2     14.00ns (1.00x)               16.00ns (1.14x)
memchr/sherlock/rare/huge2       42.4 GB/s (1.17x)             49.6 GB/s (1.00x)
memchr/sherlock/rare/small2      18.2 GB/s (1.00x)             15.1 GB/s (1.21x)
memchr/sherlock/rare/tiny2       3.1 GB/s (1.00x)              2.2 GB/s (1.38x)
memchr/sherlock/uncommon/huge2   5.8 GB/s (1.00x)              3.5 GB/s (1.63x)
memchr/sherlock/uncommon/small2  15.9 GB/s (1.00x)             5.8 GB/s (2.72x)
memchr/sherlock/uncommon/tiny2   2.9 GB/s (1.00x)              1061.3 MB/s (2.82x)
```

Ranking:

```txt
Engine                        Version  Geometric mean of speed ratios  Benchmark count
------                        -------  ------------------------------  ---------------
rust/bufchr/memchr2/prebuilt  0.0.1    1.04                            12
rust/memchr/memchr2           2.7.4    1.63                            12
```

I'll try and check the 3 needles cases afterwards because this is the use-case I am particularly interested in, then I will try integrating a proper "within-memchr" crate solution in my fork for an apples-to-apples bench with rigorous testing but after the holidays.

Yomguithereal:
Results from the 3 needles case:

```txt
benchmark                        rust/bufchr/memchr3/prebuilt  rust/memchr/memchr3
---------                        ----------------------------  -------------------
memchr/sherlock/common/huge3     2.4 GB/s (1.00x)              861.2 MB/s (2.91x)
memchr/sherlock/common/small3    6.6 GB/s (1.00x)              969.7 MB/s (7.02x)
memchr/sherlock/never/huge3      28.8 GB/s (1.93x)             55.6 GB/s (1.00x)
memchr/sherlock/never/small3     15.5 GB/s (1.43x)             22.1 GB/s (1.00x)
memchr/sherlock/never/tiny3      3.4 GB/s (1.00x)              3.4 GB/s (1.00x)
memchr/sherlock/never/empty3     14.00ns (1.00x)               16.00ns (1.14x)
memchr/sherlock/rare/huge3       27.2 GB/s (1.40x)             38.1 GB/s (1.00x)
memchr/sherlock/rare/small3      12.6 GB/s (1.07x)             13.4 GB/s (1.00x)
memchr/sherlock/rare/tiny3       3.2 GB/s (1.00x)              1935.4 MB/s (1.70x)
memchr/sherlock/uncommon/huge3   4.9 GB/s (1.00x)              2.6 GB/s (1.87x)
memchr/sherlock/uncommon/small3  12.9 GB/s (1.00x)             3.0 GB/s (4.25x)
memchr/sherlock/uncommon/tiny3   2.5 GB/s (1.00x)              626.7 MB/s (4.04x)
```

Here unrolling seems to give an edge once more on huge inputs with rare or never.

And here is the related ranking:

```txt
Engine                        Version  Geometric mean of speed ratios  Benchmark count
------                        -------  ------------------------------  ---------------
rust/bufchr/memchr3/prebuilt  0.0.1    1.13                            12
rust/memchr/memchr3           2.7.4    1.81                            12
```

BurntSushi:
Thank you for that investigation!

I do think it would be great to get this kind of optimization into this crate somehow, but it does seem like quite the challenge unfortunately.

Yomguithereal:
Thinking a little bit more about this, the only practical paradigm I can think of to implement this iteration logic all while performing runtime detection of `avx2` is callback iteration. Something like `for_each<F: FnMut>`.

PR #193, opened 2025-10-17
Avoid redudant mask computations in the generic Two and Three implementations.

adamreichold:
This does not appear to make a difference in the `rust/memchr/memchr(2|3)$` benchmarks running on x86-64 with AVX2. I would argue that it does make for simpler code though.

I suspect there is a reason for this structure and did dig through the Git history but could not find anything. (I ended up at the big "rewrite everything" commit.) At least on x86-64, I would expect fewer "movemask" instructions to be preferable due to their high latency?

### DIFF
--- a/src/arch/generic/memchr.rs
+++ b/src/arch/generic/memchr.rs
@@ -528,16 +528,16 @@ impl<V: Vector> Two<V> {
                 let eqb1 = self.v1.cmpeq(b);
                 let eqa2 = self.v2.cmpeq(a);
                 let eqb2 = self.v2.cmpeq(b);
-                let or1 = eqa1.or(eqb1);
-                let or2 = eqa2.or(eqb2);
+                let or1 = eqa1.or(eqa2);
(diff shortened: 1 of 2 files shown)


adamreichold:
Sorry for forgetting to run the tests before pushing. I obviously misunderstood something...

adamreichold:
Ok, mixed up the order in the reverse find routine. Will rerun the benchmarks to see if anything changed now...

BurntSushi:
All righty. I'll try to run these when I get chance. I have machines on my LAN specifically dedicated to benchmarking that are quiet.

> > But when matches are common, this change seems to regress runtime.
> 
> So I think why I am here is basically that I do not understand, how doing more work (extra ors and extra movemasks) can result in faster code. Especially since I do not see additional register dependencies appearing either.

I dunno. The Assembly generated might provide a clue. And if the code generated reflects the elimination of the redundant movemasks and it's still slower, then we might be looking at CPU effects that are beyond my comprehension.

PR #196, opened 2025-12-01
Hint to the compiler that returned indices are in-bounds

### LINKED ISSUE #195: Consider hinting to the compiler that indices returned from `memchr()` are within bounds of the haystack
Currently, indexing into the haystack with the index returned by memchr usually generates a bounds check, because the compiler doesn't know that it's valid. However, if an `assert_unchecked(i < haystack.len())` is inserted, the compiler has enough information to elide the check ([godbolt](https://rust.godbolt.org/z/rs9zraro4)).

Pros:
* Reduces bounds-checking in downstream code; this doesn't remove a ton of overhead but could possibly allow for further optimizations that had been inhibited by the branch.

Cons:
* `assert_unchecked()` is only available since Rust 1.81, so would either have to feature-gate this, do rust version detection, or (most likely) use `if !cond { unreachable_unchecked(); }` to polyfill it. The last option should be equivalent to `assert_unchecked()`, but I'm not sure if it's treated exactly the same by the compiler.
* Returning an out-of-bound index becomes UB rather than just a logic error. I assume there was already downstream code relying on this, though, and also it'd probably already be UB internally.

coolreader18:
Closes #195

With this change, this function in a downstream crate:
```rust
pub fn foo(x: &[u8], y: u8) -> &[u8] {
    match memchr::memchr(y, x) {
        Some(i) => &x[i..],
        None => &[],
    }
}
```

Generates this assembly:
<details>
<summary><b>Assembly output</b></summary>

```asm
foobar::foo:
	push r14
	push rbx
	push rax
	mov eax, edx
	mov rbx, rsi
	mov r14, rdi
	lea rdx, [rdi + rsi]
	mov rcx, qword ptr [rip + memchr::arch::x86_64::memchr::memchr_raw::FN@GOTPCREL]
	mov rcx, qword ptr [rcx]
	mov edi, eax
	mov rsi, r14
	call rcx
	mov rcx, rax
	test cl, 1
	je .LBB0_1
	mov rax, rdx
	sub r14, rdx
	add rbx, r14
	jmp .LBB0_2
.LBB0_1:
	mov eax, 1
	xor ebx, ebx
.LBB0_2:
	mov rdx, rbx
	add rsp, 8
	pop rbx
	pop r14
	ret
```
</details>

Whereas before, it would include a branch to `core::slice::index::slice_index_fail`.

This only works for the oneshot functions, not `memchr_iter`, since `Iter` doesn't hold a reference to the original slice. To get the same effect there, I think you'd have to add a `original_len: usize` field to `Iter`, which isn't that crazy but maybe seems silly to only use it for an optimization hint.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -234,7 +234,7 @@ jobs:
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
-        toolchain: 1.61.0
+        toolchain: 1.81.0
     - name: Basic build
       run: cargo build --verbose
     - name: Build docs
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ keywords = ["memchr", "memmem", "substring", "find", "search"]
 license = "Unlicense OR MIT"
 exclude = ["/.github", "/benchmarks", "/fuzz", "/scripts", "/tmp"]
 edition = "2021"
-rust-version = "1.61"
+rust-version = "1.81"
 
 [lib]
 name = "memchr"
--- a/src/arch/generic/memchr.rs
+++ b/src/arch/generic/memchr.rs
@@ -1132,7 +1132,15 @@ pub(crate) unsafe fn search_slice_with_raw(
     let start = haystack.as_ptr();
     let end = start.add(haystack.len());
     let found = find_raw(start, end)?;
-    Some(found.distance(start))
+    let result = found.distance(start);
+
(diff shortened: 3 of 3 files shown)


coolreader18:
Bump on this?

coolreader18:
Ooh, some recent prior (current?) art in tokio: https://github.com/tokio-rs/tokio/pull/7792

xtqqczze:
@BurntSushi PTAL

PR #204, opened 2026-02-12
lint: allow identity_op

allan2:
This adds a `lints.clippy` section to Cargo.toml, which should be supported by Rust 1.74+ and ignored by older versions.

This is related to #199.


BurntSushi:
I don't use Clippy for my open source projects.

I hope to some day, but I need to sit down and go through which lints I want. And I generally overall find it quite noisy.

PR #205, opened 2026-02-12
lint: fix clippy get_first

allan2:
This fixes [get_first](https://rust-lang.github.io/rust-clippy/master/index.html#get_first).

This is another style lint. Feel free to close if you prefer `get(0)`.


BurntSushi (approved):
Yeah I very slightly prefer `slice::first()` here so I'll bring this in. But I generally do not think this kind of a change is a good use of time. `slice::get(0)` is also perfectly clear.

Issue #206, opened 2026-02-24
[feat request] shufti table approach


hsqStephenZhang:
## Background 

Hi, I had built a crate https://crates.io/crates/shufti-matcher, which could be used to boost the case where we need to match needles whose size is greater than 3 chars. 

It would construct the [`shufti table`](https://news.ycombinator.com/item?id=44887643) at **compile** time using proc-macro so there is little overhead. 
In its impl, it used classical low&high nimble and table lookup to extract the bitmask, which could be used to locate the first char in the haystack that matches the needle. And the performance is quiet stable.

[simdjson](https://github.com/simd-lite/simd-json/blob/685257ad7073e02dfec836f91fe0eb077b434435/src/impls/neon/stage1.rs#L111) and sonic-rs adopted the shufti table approach, but made the table manually. In fact, it's quite simple to build a char set for less than 9, since there are 8 bits in a u8. For more than 8 chars, it's still possible to fit multiple chars in a bucket. 

the usage looks like:

```rust
use shufti_macro::ShuftiMatcher;
use shufti_matcher::ShuftiMatch;

#[derive(ShuftiMatcher)]
#[shufti(set = "\t\r\n")]
pub struct WsMatcher;

// https://github.com/servo/rust-url/blob/a66f4220895c3cc84ae623c218466710eb3a812f/url/src/host.rs#L131
#[derive(ShuftiMatcher)]
#[shufti(set = "\0\t\n\r #/:<>?@[\\]^|")]
pub struct MyMatcher;

fn main() {
    let s = b"\nhello\nworld";
    let pos = WsMatcher::find_first(s);
    assert_eq!(pos, Some(0));
    let pos = WsMatcher::find_first(&s[1..]);
    assert_eq!(pos, Some(5));

    let s1 = b"\nhello";
    assert_eq!(MyMatcher::find_first(s1), Some(0));

    let s2 = b"user@domain";
    assert_eq!(MyMatcher::find_first(s2), Some(4));

    let s3 = b"path\\to";
    assert_eq!(MyMatcher::find_first(s3), Some(4));

    let s4 = b"rust|cpp";
    assert_eq!(MyMatcher::find_first(s4), Some(4));

    let s5 = b"ABCdef123";
    assert_eq!(MyMatcher::find_first(s5), None);

    let s6 = b"null\0byte";
    assert_eq!(MyMatcher::find_first(s6), Some(4));

    let s7 = b" \t ";
[...]

PR #208, opened 2026-02-28
[POC] feat: memchr-n

hsqStephenZhang:
## What's this PR about? 

Implement shufti table based approach mentioned in https://github.com/BurntSushi/memchr/issues/206 to support memchr-N (N could be arbitrary usize as long as shufti algo allows)

## Background 

see https://github.com/BurntSushi/memchr/issues/206.

## Implementation

1. port `shuffle_bytes` and `shift_8bit_lane_right` methods from `aho_corasick`.
2. add `memchrn_iter, MemchrN` public API

This is still experimental, and it adds many overhead for maintenance. Perhaps a better approach is to create a separate crate named `memchr-n` to support this feature.

The purpose of this PR(now) is mainly for discussion since its quite a big change. Comments are welcomed. (But don't get trapped in current code style)

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -38,6 +38,12 @@ test = false
 doc = false
 
 [[bin]]
+name = "memchrn"
+path = "fuzz_targets/memchrn.rs"
+test = false
+doc = false
+
+[[bin]]
 name = "memrchr"
 path = "fuzz_targets/memrchr.rs"
 test = false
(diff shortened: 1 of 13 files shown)


hsqStephenZhang:
if we wrap `memchr-n` over `aho_corasick`'s Searcher, it works, but has some penalty for unnecessary abstraction.

hsqStephenZhang:
@BurntSushi Would you like to take a brief look? Do you think this approach is feasible? If so, I could fix the Ci problems and go on work on this PR.

PR #210, opened 2026-04-06
Hash-pin all actions, drop persisted credentials

woodruffw:
These are both pretty minor in your case (not a ton of exposure), but are nice-to-have in terms of making the CI more hermetic (and not dropping any credentials onto disk, even a read-only one in this case) 🙂

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: semiannually
+    open-pull-requests-limit: 1
+    cooldown:
+      default-days: 30
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,9 +89,11 @@ jobs:
           target: s390x-unknown-linux-gnu
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      with:
+        persist-credentials: false
     - name: Install Rust
-      uses: dtolnay/rust-toolchain@master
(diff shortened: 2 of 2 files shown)


BurntSushi:
My main concern here is that the CI actions don't get auto-updated. And I don't pin actions in any of my other projects. So why this one?

I guess, what specific problem does this address? What happens if I don't mergr this? I use few actions and only ones from places that I trust.

woodruffw:
> My main concern here is that the CI actions don't get auto-updated. And I don't pin actions in any of my other projects. So why this one?

If you're open to it, I can add a Dependabot or similar config that'll keep the actions auto-updated. That's what I do personally.

(Note that you can set Dependabot to be minimally noisy, e.g. update only twice a year and combine all action bumps into a single PR.)

Re: other projects: I was actually slowly burning these down, and was going to send more PRs for the other ones. More generally I'm working backwards through uv's dep tree 😅 



> I guess, what specific problem does this address? What happens if I don't mergr this? I use few actions and only ones from places that I trust.

The specific risk is the kind of "mutable pointer via tags" that happened with the [Trivy attack](https://www.wiz.io/blog/trivy-compromised-teampcp-supply-chain-attack). 

In your case the risk is pretty marginal though -- I think you're not doing any publishing via CI so the risk of an attacker compromising these actions is mostly limited to them being a nuisance. So if the added indirection here is unappealing to you then this is safe to close.

BurntSushi:
I might some day set up trusted publishing. Or it at least seems like a good idea.

I think I'll take you up on the dependendabot offer. Specifically, i like the minimally noisy option of twice a year. Thank you. :-)

woodruffw:
Done! [`4b3e7d8` (#210)](https://github.com/BurntSushi/memchr/pull/210/commits/4b3e7d8aa58c0a7f0e3d4fad5b4f473bbfc46f3e) will send an update PR twice a year, with a 30 day cooldown for any updates therein.

BurntSushi (approved):
Let's give it a go!

PR #215, opened 2026-05-01
Fix minor UB flagged by miri

Shnatsel:
Fixed UB in the generic packed-pair finder when the safe `Finder::find` API is called with a search-time needle longer than the haystack.

Although callers are documented to pass the same needle used at construction time, this is still a safe API. A mismatched longer needle could make `find_in_chunk` evaluate `end.sub(needle.len())`, creating a pointer before the start of the haystack allocation. That pointer was only used for comparison, not dereferenced, so ASan does not catch it, but Miri reports it as UB.

The fix checks the remaining byte length from the candidate pointer to `end` before forming any pointer derived from `end`. If the search-time needle cannot fit, the function now returns `None` without invalid pointer arithmetic. This is verified with a Miri regression test.

### DIFF
--- a/src/arch/generic/packedpair.rs
+++ b/src/arch/generic/packedpair.rs
@@ -223,8 +223,9 @@ impl<V: Vector> Finder<V> {
     /// # Safety
     ///
     /// It must be safe to do an unaligned read of size(V) bytes starting at
-    /// both (cur + self.index1) and (cur + self.index2). It must also be safe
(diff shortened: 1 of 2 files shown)


xtqqczze:
@BurntSushi Completed by https://github.com/BurntSushi/memchr/pull/226.

BurntSushi:
Ah apologies that I either didn't see this PR or forgot about it. Thank you!

PR #216, opened 2026-05-19
chore: improve memchr maintenance path

vip892766gma:
Summary:
- Add additional edge-case unit tests in the existing #[cfg(test)] modules covering empty-needle, single-byte haystack, and boundary alignment cases for the generic (all-arch) memchr/packedpair fallback paths, plus a debug_assert for needle length invariants at one public entry point where the invariant is currently only implicit. Pure additive tests and one narrow assertion — no behavior change.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/src/arch/all/memchr.rs
+++ b/src/arch/all/memchr.rs
@@ -1019,4 +1019,72 @@ mod tests {
         let data = [0, 0, 0, 0, 0, 0, 0, 0];
         assert_eq!(One::new(b'\x00').find(&data), Some(0));
     }
+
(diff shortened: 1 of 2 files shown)


xtqqczze:
@BurntSushi For context: https://github.com/rust-lang/rust/pull/156958#issuecomment-4544561954

The Rust moderation team suspended the contributor over repeated submission of insufficiently reviewed LLM-generated PRs.

PR #220, opened 2026-05-24
Fix NeonMoveMask and align naming with actual semantics

### LINKED ISSUE #219: Possible bug in `NeonMoveMask`
On [line 390 of vector.rs](https://github.com/BurntSushi/memchr/blob/master/src/vector.rs#L390), am I hallucinating or should:
```rust
((1 << n) << 2)
```
be instead:
```rust
(1 << (n << 2))
```
?

(Though honestly I'd just replace the whole thing with `NeonMoveMask(0x8888_8888_8888_8888 << (n << 2))` so that the returned mask is actually valid, neon-wise.)

HansBrende:
Fixes #219 

I also updated the naming to align with the semantics, given by the comment "The mask has all of its bits **set** except for the first N least significant bits", and saved a few bit ops.

### DIFF
--- a/src/arch/generic/packedpair.rs
+++ b/src/arch/generic/packedpair.rs
@@ -89,7 +89,7 @@ impl<V: Vector> Finder<V> {
             haystack.len(),
         );
 
-        let all = V::Mask::all_zeros_except_least_significant(0);
+        let all = V::Mask::all_ones_except_least_significant(0);
         let start = haystack.as_ptr();
         let end = start.add(haystack.len());
         let max = end.sub(self.min_haystack_len);
@@ -139,7 +139,7 @@ impl<V: Vector> Finder<V> {
             // significant bits, where N=overlap. This way, any matches that
             // occur in find_in_chunk within the overlap are automatically
             // ignored.
-            let mask = V::Mask::all_zeros_except_least_significant(overlap);
+            let mask = V::Mask::all_ones_except_least_significant(overlap);
             cur = max;
             let m = self.find_in_chunk(needle, cur, end, mask);
             if let Some(chunki) = m {
--- a/src/tests/mod.rs
+++ b/src/tests/mod.rs
@@ -3,6 +3,7 @@ pub(crate) mod memchr;
 pub(crate) mod packedpair;
 #[macro_use]
 pub(crate) mod substring;
+pub(crate) mod vector;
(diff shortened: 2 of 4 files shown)


BurntSushi:
There is no failing test before or after this change, which makes me suspicious of test coverage. Could you please add a regression test first? Thanks.

HansBrende:
@BurntSushi I don't think that's possible... as the bug would have only manifested in duplicated work, not wrong outputs. As far as I can tell. (Essentially the mask would have extra 1s in it, causing unnecessary rechecks at the end of the find loop.)

BurntSushi:
OK, that means there is a gap in my understanding here. I'll need to dig into this. If it's not causing any actual change in behavior, then this will probably be low priority for me.

HansBrende:
@BurntSushi hopefully I can help with that understanding gap:

Current behavior: 
```rust
fn all_zeros_except_least_significant(n: usize) -> NeonMoveMask {
    NeonMoveMask(!(((1 << n) << 2) - 1))
    
    // which is the same as:
    NeonMoveMask((0xFFFF_FFFF_FFFF_FFFF << n) << 2)
    
    // which is the same as:
    NeonMoveMask(0xFFFF_FFFF_FFFF_FFFF << (n + 2))
}
```

Why this is incorrect:

it should be: 

```rust
0xFFFF_FFFF_FFFF_FFFF << (n*4)
```
Because for the NeonMoveMask "hack", each bit index is spaced 4 bits from the previous bit index. So the offset needs to simply be multiplied by 4.

i.e., `0xFFFF_FFFF_FFFF_FFFF << (n << 2)`, **not** `(0xFFFF_FFFF_FFFF_FFFF << n) << 2`

But better is `0x8888_8888_8888_8888 << (n << 2)` so that the resulting mask is actually valid (bit positions at 0, 1, and 2 mod 4 are supposed to be 0).

Does this help?

HansBrende:
@BurntSushi I have now added a test that will fail badly in the existing implementation (although not tied directly to the public-facing API). Hopefully that helps as well.

HansBrende:
@BurntSushi in order to demonstrate those test failures for the existing codebase, I opened PR #221  with the same test-cases added but my change reverted. Please let me know once you have a chance to look at the test failures so I can close that PR.

PR #221, opened 2026-05-24
Show test failures for 219

HansBrende:
I'm only opening this PR to show the test failures when I revert my change from #220. Will close after the tests complete.


HansBrende:
Oh apparently the test results disappear when I close a PR 😑

Guess I will keep it open until the test results aren't needed anymore.

PR #228, opened 2026-07-06
let compiler know that indexes returned by memchr fns are in bounds

### LINKED ISSUE #195: Consider hinting to the compiler that indices returned from `memchr()` are within bounds of the haystack

Dr-Emann:
For example, in code like:

```rust
let len = memchr::memchr(b'\0', slice).unwrap_or(slice.len());
&slice[..len]
```

Without this change, there is a potential panic at `&slice[..len]`, but with this change, rust knows the len must be valid, and no longer adds the potential panic.

Same as #196, but without needing to bump MSRV

Closes #196
Closes #195

### DIFF
--- a/src/arch/generic/memchr.rs
+++ b/src/arch/generic/memchr.rs
@@ -1132,7 +1132,13 @@ pub(crate) unsafe fn search_slice_with_raw(
     let start = haystack.as_ptr();
     let end = start.add(haystack.len());
     let found = find_raw(start, end)?;
-    Some(found.distance(start))
+    let idx = found.distance(start);
+    // Required by safety invariant required for find_raw
+    // this lets the compiler know the returned index is in bounds for the slice
+    if idx >= haystack.len() {
+        core::hint::unreachable_unchecked();
+    }
+    Some(idx)
 }
 
 /// Performs a forward byte-at-a-time loop until either `ptr >= end_ptr` or


Dr-Emann:
Confirmed by adding a git-based dependency (`memchr = { version = "2.8.2", git = "https://github.com/Dr-Emann/memchr.git", branch = "push-pwmovxxtxtsn" }`) that a function like:

```rust
#[inline(never)]
pub fn do_slice(slice: &[u8]) -> &[u8] {
    let len = memchr::memchr(b'\0', slice).unwrap_or(slice.len());
    &slice[..len]
}
```

no longer has a possible panic anymore, but did before.

BurntSushi:
Ah interesting, I was wrong then about this potentially not working. I did that test myself. The codegen before (on x86-64):

```
.section .text.memchr_inline_assertion::do_slice,"ax",@progbits
	.p2align	4
.type	memchr_inline_assertion::do_slice,@function
memchr_inline_assertion::do_slice:
	.cfi_startproc
	push r14
	.cfi_def_cfa_offset 16
	push rbx
	.cfi_def_cfa_offset 24
	push rax
	.cfi_def_cfa_offset 32
	.cfi_offset rbx, -24
	.cfi_offset r14, -16
	mov rbx, rsi
	mov r14, rdi
	lea rdx, [rdi + rsi]
	mov rax, qword ptr [rip + memchr::arch::x86_64::memchr::memchr_raw::FN@GOTPCREL]
	mov rax, qword ptr [rax]
	xor edi, edi
	mov rsi, r14
	call rax
	cmp rax, 1
	jne .LBB5_3
	sub rdx, r14
	cmp rdx, rbx
	ja .LBB5_4
	mov rbx, rdx
.LBB5_3:
	mov rax, r14
	mov rdx, rbx
	add rsp, 8
	.cfi_def_cfa_offset 24
	pop rbx
	.cfi_def_cfa_offset 16
	pop r14
	.cfi_def_cfa_offset 8
	ret
.LBB5_4:
	.cfi_def_cfa_offset 32
	lea rcx, [rip + .Lanon.5532cfe719cce4bb48693cad70756471.5]
	xor edi, edi
	mov rsi, rdx
	mov rdx, rbx
	call qword ptr [rip + core::slice::index::slice_index_fail@GOTPCREL]
```

And the codegen afterward:

```
.section .text.memchr_inline_assertion::do_slice,"ax",@progbits
	.p2align	4
.type	memchr_inline_assertion::do_slice,@function
memchr_inline_assertion::do_slice:
	.cfi_startproc
	push r14
	.cfi_def_cfa_offset 16
	push rbx
	.cfi_def_cfa_offset 24
	push rax
	.cfi_def_cfa_offset 32
	.cfi_offset rbx, -24
	.cfi_offset r14, -16
	mov r14, rsi
	mov rbx, rdi
	lea rdx, [rdi + rsi]
	mov rax, qword ptr [rip + memchr::arch::x86_64::memchr::memchr_raw::FN@GOTPCREL]
	mov rax, qword ptr [rax]
	xor edi, edi
	mov rsi, rbx
	call rax
	cmp rax, 1
	jne .LBB5_2
	sub rdx, rbx
	mov r14, rdx
.LBB5_2:
	mov rax, rbx
	mov rdx, r14
	add rsp, 8
	.cfi_def_cfa_offset 24
	pop rbx
	.cfi_def_cfa_offset 16
	pop r14
	.cfi_def_cfa_offset 8
	ret
```

That's a nice improvement!

BurntSushi:
This PR is on crates.io in `memchr 2.8.3`.

PR #230, opened 2026-07-11
Allow the libc engine to build on msvc targets

ChrisDenton:
@tgross35 noted that the libc engine had to be patched in order to build the other libc benchmarks on Windows msvc. So this PR changes the missing `libc::memmem` compile time error into a runtime error.


BurntSushi (approved):
Nice thank you!

PR #235, opened 2026-07-28
cargo: remove deprecated package authors field

xtqqczze:
`package.authors` is deprecated: https://github.com/rust-lang/cargo/pull/15068


BurntSushi:
I don't see a good reason to remove this

Issue #236, opened 2026-07-30
Add AVX-512 support


BurntSushi:
It would be great to add support to this for `memchr`.

This was previously attempted in https://github.com/BurntSushi/memchr/pull/169 where I was skeptical, but it seems like AVX-512 is here to stay given that now even AMD CPUs ship with support for them. It [seems like downclocking isn't as big of an issue these days](https://chipsandcheese.com/p/zen-5s-avx-512-frequency-behavior), which was one of my main concerns.

PR #239, opened 2026-08-12
Implementation of AVX-512

pedrodesu:
Implemented AVX-512. Also naturally had to adapt `vector.rs` and change vector byte limit to 64. Some a bit off-topic but rather small and simple changes, made a change to `is_available` which, in my opinion, could be made a bit simpler, albeit functionally equal; Changed "Used for haystacks bigger than N bytes." to "Used for haystacks of N bytes or more." (as N bytes is enough to use the respective strategy but the former implies that it must be bigger than N, which is incorrect); Also made a small change on the safety comments of `find` and `find_prefilter` inside of `avx2/packedpair.rs`. These mention calling sse2 routines where we're calling avx2 routines, so I believe these were overlooked mistakes.

Solves https://github.com/BurntSushi/memchr/issues/236.

Let me know your thoughts.
Thank you so much for this lib!


pedrodesu:
The AVX-512 intrinsic types were nightly-only behind `#![feature(stdsimd)]` in Rust 1.61.0, the MSRV of the library as of now. They were not stabilized until Rust 1.65.0. As such, it's impossible to support stable compilation of AVX-512 code paths on toolchains prior to this version without feature-gating AVX-512 out entirely for older compiler versions.

I propose raising MSRV to at least 1.65.0 if the maintainers agree. Otherwise I don't think this is a possibility.

